PR #10, opened 2024-12-09
Create datamodel base classes based on openbisschema

### LINKED ISSUE #3: Create datamodel base classes based on OpenbisSchema
This is a super-issue for tracking the Python definition of our data classes. It requires a lot of testing and use cases, so this description will be edited to add more sub-tasks when we find out what we need to do.

JosePizarro3:
Closes #3

### DIFF
--- a/bam_data_store/metadata/definitions.py
+++ b/bam_data_store/metadata/definitions.py
@@ -1,0 +1,423 @@
+import re
+from enum import Enum
+from typing import Any, Optional
+
+from pydantic import BaseModel, Field, field_validator, model_validator
+
+
+class DataType(str, Enum):
+    """Enumeration of the data types available in openBIS."""
+
+    BOOLEAN = 'BOOLEAN'
+    CONTROLLEDVOCABULARY = 'CONTROLLEDVOCABULARY'
+    DATE = 'DATE'
+    HYPERLINK = 'HYPERLINK'
+    INTEGER = 'INTEGER'
+    MATERIAL = 'MATERIAL'
+    MULTILINE_VARCHAR = 'MULTILINE_VARCHAR'
+    OBJECT = 'OBJECT'
+    REAL = 'REAL'
+    TIMESTAMP = 'TIMESTAMP'
+    VARCHAR = 'VARCHAR'
+    XML = 'XML'
+
+    @property
+    def pytype(self) -> type:
+        """
+        Maps the openBIS data type to its corresponding Python type.
+
+        Returns:
+            type: The native Python type of the openBIS data type.
+        """
+        # TODO check the other data types
+        mapping = {
+            'BOOLEAN': bool,
+            # 'CONTROLLEDVOCABULARY': ,
+            # 'DATE': ,
+            'HYPERLINK': str,
+            'INTEGER': int,
+            # 'MATERIAL': ,
+            'MULTILINE_VARCHAR': str,
+            # 'OBJECT': ,
+            'REAL': float,
+            # 'TIMESTAMP': ,
+            'VARCHAR': str,
+            # 'XML': ,
+        }
+        return mapping.get(self, None)
+
+
+class EntityDef(BaseModel):
+    """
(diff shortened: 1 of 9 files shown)


bam_data_store/metadata/definitions.py
```diff
@@ -0,0 +1,428 @@
...
+    Definition class used for the common fields for `CollectionTypeDef`, `ObjectTypeDef`, and `DataSetType`.
+    It adds the fields of `validation_script`.
+    """
+
+    validation_script: Optional[str] = Field(
+        default=None,
+        description="""
+        Script written in Jython used to validate the object type.
+        """,
+    )  # ? is this truly used?
```
carlosmada22, line 115: Never seen this "BaseObject" creation (check also with Angela). I think it is possible to move this validation_script to "EntityDef", and make "ObjectTypeDef" inherit from EntityDef directly. If not possible, leave it like it is, but just for the validation_script (I think this will not be created directly)

bam_data_store/metadata/definitions.py
```diff
@@ -0,0 +1,428 @@
...
+        default=None,
+        description="""""",
+    )
+
+    dynamic_script: Optional[str] = Field(
+        default=None,
+        description="""""",
+    )
+
+    multivalued: Optional[str] = Field(
```
carlosmada22, line 288: Check if this is being used. For what I know, properties will have just one unique value.

carlosmada22 (approved):
Everything seems correct, missing the types that I will automate using scripts. I would like to review the tests together in the future, but the branch can be merged.

bam_data_store/metadata/definitions.py
```diff
@@ -0,0 +1,428 @@
...
+    description: str = Field(
+        ...,
+        description="""
+        Description of the entity. This is the human-readable text for the object and must be
+        as complete and concise as possible. The German description can be added after the English
+        description separated by a double slash (//), e.g. `'Chemical Substance//Chemische Substanz'`.
+        """,
+    )
+
+    # TODO check ontology_id, ontology_version, ontology_annotation_id, internal (found in the openBIS docu)
```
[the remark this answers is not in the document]
JosePizarro3: Ok. I re-opened a ticket in the openBIS forum (https://community.openbis.ch/t/how-can-semantic-annotations-be-used-to-describe-openbis-sample-property-types/49/6) asking about this. I will keep this comment so we don't lose track.

bam_data_store/metadata/definitions.py
JosePizarro3: I meant, the field `validation_script`.

The reason why there is a `BaseObjectTypeDef` class is that not all other entities contain a `validation_script` field, so if we put it in `EntityDef`, properties, for example, would have this field (and in openBIS, it seems they don't).

In other words, all entities contain `version`, `code` and `description` (that's why they are defined in `EntityDef`), while not all contain `validation_script` (as you can see in the inheritance, only `CollectionTypeDef`, `ObjectTypeDef` and `DataSetTypeDef` do).

bam_data_store/metadata/definitions.py
JosePizarro3: Yeah, I don't see anywhere why I added this.

JosePizarro3:
> Everything seems correct, missing the types that I will automate using scripts. I would like to review the tests together in the future, but the branch can be merged.

Ok, thanks for the review!

I think everything would make more sense with some documentation. I can dedicate myself to do this later this week, so next year will be, hopefully, ready.

bam_data_store/metadata/definitions.py
JosePizarro3: I deleted it

bam_data_store/metadata/definitions.py
JosePizarro3: I deleted now this question, as I know `validation_script` is added in openBIS; but tbh, we are going to handle this differently, instead of using Jython, we can use these classes and pydantic for validating stuff.

PR #18, opened 2024-12-17
Extract datamodel to json and excel

JosePizarro3:
Added CLI features to extract the datamodel objects and print them to json in a subfolder `bam_data_store/datamodel/artifacts/`.


JosePizarro3:
I've merged the export-to-excel implementation as well

JosePizarro3:
@carlosmada22 this is my changes to generate JSONs and the excel in the typical format of openBIS from our classes in the `datamodel/`. This is a good opportunity for you to see a better CLI than argparse (using `click`); check the `cli/` folder please.

Could you try checking out in this branch and doing:
```bash
bam_masterdata export_to_json
```
or:
```bash
bam_masterdata export_to_excel
```

And see whether it makes sense what is being printed out.

With this changes, we can start thinking about connecting excel with Python/JSON, and allow people to have a full environment for developing data models. There is left a TODO, which is in #35 (i.e., exporting/importing from WebProtege), but this we can do later.

---
And btw, I had to fix a minor thing from the `fill_masterdata.py`: apparently, the data type has changed from SAMPLE to OBJECT at some point. Now it is working.

PR #47, opened 2025-01-14
13 add push to openbis to update masterdata in openbis

JosePizarro3:


JosePizarro3:
@carlosmada22 back in the day (2.5 months ago) I did this. The important thing is the method `to_openbis` in entities.py, everything else is probably useless nowadays

PR #56, opened 2025-01-22
Excel to entities

carlosmada22:


bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
+import re
+from typing import TYPE_CHECKING, Any
+
+# import logger
```
JosePizarro3, line 4: `from bam_masterdata.logger import logger`

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
+import re
+from typing import TYPE_CHECKING, Any
+
+# import logger
+import openpyxl
+
+# logger = logger()
+
+
+def index_to_excel_column(index):
```
JosePizarro3, line 10: typing and description? I let you decide

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+
+def index_to_excel_column(index):
+    column = ""
+    while index > 0:
+        index, remainder = divmod(index - 1, 26)
+        column = chr(65 + remainder) + column
+    return column
+
+
+def get_last_non_empty_row(sheet, start_index):
```
JosePizarro3, line 18: typing? again, you decide if whether doing it


I think is:
```python
def get_last_non_empty_row(sheet: Worksheet, start_index: int) -> int:
```

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+    property_dict = {}
+    expected_terms = [
+        "Code",
+        "Description",
+        "Mandatory",
+        "Show in edit views",
+        "Section",
+        "Property label",
+        "Data type",
+        "Vocabulary code",
```
JosePizarro3, line 55: Can you add `"Object code"` commented out? This is because in the excel import system is missing, but for our Python and other stuff we need this information when the `Data type` is OBJECT

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+    ]
+
+    header_index = start_index_row + 3
+    row_headers = [cell.value for cell in sheet[header_index]]
+
+    # print("Header index:", header_index, "//Header:", row_headers)
```
JosePizarro3, line 61: delete this dead code if not needed (and also everywhere in the document)

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+    # code_index = row_headers.index("Code") + 1
+    (
```
JosePizarro3, line 63: Better to use a dictionary:
```python
    expected_terms = {
        "Code": [],
        "Description": [],
        "Mandatory": [],
        "Show in edit views": [],
        "Section": [],
        "Property label": [],
        "Data type": [],
        "Vocabulary code": [],
        "Object code": [],
}
```

no?

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+        sections,
+        labels,
+        data_types,
+        vocabulary_codes,
+    ) = [], [], [], [], [], [], [], []
+
+    for term in expected_terms:
+        if term not in row_headers:
+            if term in ("Mandatory", "Show in edit views", "Section"):
+                print(f"Warning: '{term}' not found in the properties headers.")
```
JosePizarro3, line 77:
```python
            if term in ("Mandatory", "Show in edit views", "Section"):
                logger.warning(f"'{term}' not found in the properties headers.")
            else:
                logger.error(f"'{term}' not found in the properties headers.")
```

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+            # print(term_index)
+
+            # Check the column below "Code"
+            if term == "Code":
+                for cell in sheet[term_letter][header_index:last_non_empty_row]:
+                    if cell.value is not None:
+                        codes.append(cell.value)
+                    else:
+                        codes.append(None)
+                # invalid_codes = [
```
JosePizarro3, line 93: now that I see this code, this is not dead code, but actually we should have these checks and print the messages to the logger, right?

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+def block_to_entity_dict(sheet, start_index_row, last_non_empty_row, complete_dict):
+    attributes_dict = {}
+
+    entity_type_position = f"A{start_index_row}"
+    entity_type = sheet[entity_type_position].value
+
+    entity_types = [
+        "OBJECT_TYPE",
```
JosePizarro3, line 334: Why do we need both OBJECT_TYPE and SAMPLE_TYPE here?

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+
+    header_terms = [cell.value for cell in sheet[start_index_row + 1]]
+
+    if entity_type not in entity_types:
+        print(
+            "The entity type (cell A1) should be one of the following: SAMPLE_TYPE/OBJECT_TYPE, EXPERIMENT_TYPE/COLLECTION_TYPE, DATASET_TYPE, PROPERTY_TYPE, VOCABULARY_TYPE"
+        )
+        # return "\n".join(errors)
+    else:
+        if entity_type == "SAMPLE_TYPE" or entity_type == "OBJECT_TYPE":
```
JosePizarro3, line 350: for readability, this and the previous ifs in the `terms_to_dict` could be part of a class and methods applied to each type. There you can identify common abstracted methods to be applied

bam_masterdata/cli/excel_to_entities.py
```diff
@@ -0,0 +1,661 @@
...
+        sheets_dict[normalized_sheet_name] = {}
+
+        while start_row <= sheet.max_row:
+            # Find the last non-empty row of the current block
+            last_non_empty_row = get_last_non_empty_row(sheet, start_row)
+
+            # Check if we've reached the end of the sheet or found two consecutive empty rows
+            if last_non_empty_row is None:
+                if i == len(sheet_names) - 1:  # Check if it's the last sheet
+                    print(
```
JosePizarro3, line 624: logger.info(...)

bam_masterdata/cli/cli.py
```diff
@@ -38,20 +38,44 @@ def cli():
...
+    required=False,
+    help="""
+    (Optional) The path to the Masterdata Excel file.
+    """,
+)
+def fill_masterdata(url, path):
     start_time = time.time()
 
+    # Define output directory
+    output_directory = "./artifacts/tmp/" if path else DATAMODEL_DIR
```
JosePizarro3, line 53: what about exporting to `os.path.join(DATAMODEL_DIR, "tmp")` instead of in artifacts? This way the one creating this temporary files can compare if there are already some python modules under DATAMODEL_DIR

JosePizarro3 (commented):
Very nice! can you add some testing and some excels where to do this?

Also, I added a few minor comments on minor design this you can decide if accept or just keep doing it with your style.

Then, one open point: some code doing checks on the format is commented out. We would actually like to keep this, right? Is this the issue you were having with the logger?

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Already done in the last commit

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Same, ready

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Typing will be applied in the next commit correcting this

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Added, but you mean to have it there for future versions of the Excel files that will contain also the header "Object code"?

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Resolved in the last commit uploaded too

bam_masterdata/cli/excel_to_entities.py
carlosmada22: Already done in last commit

bam_masterdata/cli/excel_to_entities.py
carlosmada22: We can discuss this

bam_masterdata/cli/cli.py
carlosmada22: Sure, do not mind where to locate the files, I have set artifacts because was the last thing we have talked about. Changed it for next commit

bam_masterdata/cli/excel_to_entities.py
JosePizarro3: well, now I am not so sure; this is something internal from openBIS, but it looks like a bug to me: we need to somehow define the code of the object for when a property is of dataType OBJECT

Otherwise, how do we know the object that an assigned property is referring to?

bam_masterdata/cli/cli.py
```diff
@@ -38,20 +38,44 @@ def cli():
...
-def fill_masterdata(url):
+@click.option(
+    "--excel-file",
+    type=click.Path(exists=True, dir_okay=False),
+    required=False,
+    help="""
+    (Optional) The path to the Masterdata Excel file.
+    """,
+)
+def fill_masterdata(url, path):
```
JosePizarro3, line 49: you have to change path to excel_file in this function

maybe click has another option for aliases if you prefer so

JosePizarro3:
Ok, I went locally through the branch and I got some points:

- We need to change all `print` for logger messages. The level of the messages has to be clear.
- I am having some errors when a cell.value is TRUE or FALSE. In my case, `openpyxl`  is correctly reading it as a boolean, while I think for you, @carlosmada22, it is a string. My openpyxl version is 3.1.5.
- Some of the logic is repeated between the ifs, e.g., the case when a cell.value is TRUE or FALSE. This can go in its own function to avoid repetition.

PR #61, opened 2025-01-28
Fixing mypy and other extras

JosePizarro3:
I added a function `str_to_bool` to leverage out some logic from the `excel_to_entities.py` module.

I also added the `tmp/` folder which is created to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,5 @@ pytest-coverage.txt
 
 # artifacts
 artifacts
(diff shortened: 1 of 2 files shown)


JosePizarro3:
@carlosmada22 this is my fix on mypy

I am having conflicts because I rebased this branch with respect to `main`, so if you rebase in `excel_to_entities`, the conflicts will get fixed

PR #70, opened 2025-01-31
63 clean code and refactor into class   excel to entities

carlosmada22:
This pull request includes changes to the `.vscode/settings.json` file to add a new debug configuration, and modifications to the `bam_masterdata` module to refactor the way Excel data is extracted. The most important changes are summarized below:

### Debug Configuration Improvements:

* [`.vscode/settings.json`](diffhunk://#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357R36): Added a new debug configuration for exporting to RDF and set `justMyCode` to `false` to include library code in debugging. [[1]](diffhunk://#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357R36) [[2]](diffhunk://#diff-a5de3e5871ffcc383a2294845bd3df25d3eeff6c29ad46e3a396577c413bf357R84-R95)

### Code Refactoring:

* [`bam_masterdata/cli/fill_masterdata.py`](diffhunk://#diff-7348b5d5d645b7711563e3f802d6bba9ad4105c6bed7cc58f5c840b2a18bc0e4L5-R5): Refactored to use `MasterdataExcelExtractor` for extracting entities from Excel, replacing the previous `excel_to_entities` function. [[1]](diffhunk://#diff-7348b5d5d645b7711563e3f802d6bba9ad4105c6bed7cc58f5c840b2a18bc0e4L5-R5) [[2]](diffhunk://#diff-7348b5d5d645b7711563e3f802d6bba9ad4105c6bed7cc58f5c840b2a18bc0e4L30-R32)
* [`bam_masterdata/excel/__init__.py`](diffhunk://#diff-4a801fe2ea1f913a97a7b681964e7c644be4aaa51628daffb05cee586015d15fR1): Added import for `MasterdataExcelExtractor` to the module's `__init__.py` file.


bam_masterdata/excel/excel_to_entities.py
```diff
@@ -0,0 +1,701 @@
...
+from typing import TYPE_CHECKING, Any, Optional, Union
+
+import openpyxl
+from openpyxl.worksheet.worksheet import Worksheet
+
+from bam_masterdata.logger import logger
+
+
+class MasterdataExcelExtractor:
+    DATA_TYPES = [
```
JosePizarro3, line 11: You can use the Enum defined in `bam_masterdata.metadata.definitions import DataType`. If you want to check if some `value` is allowed by the data types, you just need to do:
```python
value in [dt.value for dt in DataType]
```

bam_masterdata/excel/excel_to_entities.py
```diff
@@ -0,0 +1,701 @@
...
+        "BOOLEAN",
+        "CONTROLLEDVOCABULARY",
+        "XML",
+        "TIMESTAMP",
+        "DATE",
+        "SAMPLE",
+        "OBJECT",
+    ]
+
+    VALIDATION_RULES = {
```
JosePizarro3, line 26: Can you add a todo comment before this variable?

```python
# TODO move these validation rules to a separate json
```

This is because I think we could have these mappings in a separate JSON to make it more readable. But you can leave it for another pr :)

bam_masterdata/excel/excel_to_entities.py
```diff
@@ -0,0 +1,701 @@
+import re
+from typing import TYPE_CHECKING, Any, Optional, Union
+
+import openpyxl
+from openpyxl.worksheet.worksheet import Worksheet
```
JosePizarro3, line 5: Ok, I checked the code and it seems `Worksheet` is only used for typing. You can:
```python
import re
from typing import TYPE_CHECKING, Any, Optional, Union

if TYPE_CHECKING:
    from openpyxl.worksheet.worksheet import Worksheet
```

Then, you need to change in all the times a `Worksheet` was typed in a function by `"Worksheet"`, i.e., add the double-quotes, e.g.:
```python
    def get_last_non_empty_row(
        self, sheet: "Worksheet", start_index: int
    ) -> Optional[int]:
```

The thing is that, if a class is only used for typing checks, then it does not need to be imported (and hence added to memory) every time the module is executed. That's why we have this TYPE_CHECKING and we just need to add an if condition + double quotes when typing the specific class.

JosePizarro3 (approved):
Looks perfect, congrats! 🙂 

Just a couple of comments for you to discuss with me if you agree or not.

Maybe it is a good idea to add testing for this class, but this I leave for you to decide if you want to do it in this PR or open a new issue and tackle it later.

carlosmada22:
> Maybe it is a good idea to add testing for this class, but this I leave for you to decide if you want to do it in this PR or open a new issue and tackle it later.

Perfect, I will open a new issue for the testing after this pull request.

bam_masterdata/excel/excel_to_entities.py
carlosmada22: Changed thank you!

PR #72, opened 2025-02-03
Validation mappings moved to excel and logic for reading it in Extractor implemented

carlosmada22:
Currently, validation mappings are hardcoded within the checker, making it difficult to modify or extend them without code changes. To improve maintainability and flexibility, the mappings should be extracted into a separate JSON file that the checker can load dynamically.

**Changes:**

- Create a structured JSON file to store validation mappings.
- Modify the checker to read and apply validation rules from the external JSON.
- Ensure proper error handling in case the JSON is missing or malformed.

**Expected Benefits:**

- Easier updates to validation rules without modifying code.
- Improved readability and maintainability.
- Potential for reusability across multiple components.

**Additional Notes:**

- Added a fallback mechanism in case the JSON file is unavailable.
- Validate JSON structure before use to prevent runtime issues.


bam_masterdata/excel/excel_to_entities.py
```diff
@@ -12,125 +14,56 @@
...
         self.excel_path = excel_path
         self.workbook = openpyxl.load_workbook(excel_path)
         self.logger = logger
 
+        # Load validation rules at initialization
+        if not MasterdataExcelExtractor.VALIDATION_RULES:
+            self.load_validation_rules()
+
+    @classmethod
+    def load_validation_rules(cls, file_path: str = None):
```
JosePizarro3, line 30: I think this function is not only for `MasterdataExcelExtractor`. I will take it out, probably to utils or to a separate folder of the CLI (maybe we want something like `bam_masterdata/validation_rules/`), and put this function there.

Then you just need to change the lines where `self.VALIDATION_RULES` is called.

bam_masterdata/excel/excel_to_entities.py
```diff
@@ -12,125 +14,56 @@
...
+        """Loads validation rules from a JSON file."""
+        if file_path is None:
```
JosePizarro3, line 32: mypy will probably complain because you considered `file_path` to be of type `str`, but it is actually `Optional[str]`

You can simplify these lines as well, instead of None you can point to a specific validation rule JSON. Like `"./bam_masterdata/....."`

bam_masterdata/excel/excel_to_entities.py
```diff
@@ -12,125 +14,56 @@
...
+        if not os.path.exists(file_path):
+            logger.error(f"Validation rules file not found: {file_path}")
+            raise FileNotFoundError(f"Validation rules file not found: {file_path}")
+
+        try:
+            with open(file_path, encoding="utf-8") as file:
+                cls.VALIDATION_RULES = json.load(file)
+
+            # Validate JSON structure
+            if not isinstance(cls.VALIDATION_RULES, dict):
```
JosePizarro3, line 48: I think this is not necessary. If json.load(file) does not return a dict, for sure it will show an error

JosePizarro3 (commented):
Just an idea for moving to a different folder the validation rules and how to call for them

PR #80, opened 2025-02-10
Fix property assignment setattr

JosePizarro3:
This pull request includes several changes to the `bam_masterdata` package, focusing on metadata definitions, entity handling, and test coverage. The most important changes include the addition of datetime imports, updates to the `DataType` class, enhancements to the `BaseEntity` class, and corresponding test updates.

### Metadata Definitions and Types:

* [`bam_masterdata/metadata/definitions.py`](diffhunk://#diff-2932caf9fee85eb8a9c3b6c4c9b3116686cdeca6cc3ccb74a51ef9f093e4b6d9R1): Added `datetime` import and updated `DataType` class to deprecate `MATERIAL` and include `DATE` and `TIMESTAMP` in the `pytype` method with their respective datetime types. [[1]](diffhunk://#diff-2932caf9fee85eb8a9c3b6c4c9b3116686cdeca6cc3ccb74a51ef9f093e4b6d9R1) [[2]](diffhunk://#diff-2932caf9fee85eb8a9c3b6c4c9b3116686cdeca6cc3ccb74a51ef9f093e4b6d9L18-R19) [[3]](diffhunk://#diff-2932caf9fee85eb8a9c3b6c4c9b3116686cdeca6cc3ccb74a51ef9f093e4b6d9L38-R45)

### Entity Handling:

* [`bam_masterdata/metadata/entities.py`](diffhunk://#diff-e8d59d232a103d745395193665c31c153d84d50db8e0ec82737b01e7dcbda865R27-R64): Enhanced `BaseEntity` class by adding an `__init__` method to store `_property_metadata`, implementing `__setattr__` for type checking, and adding a `__repr__` method for better string representation. Introduced `set_property_metadata` method to store property metadata. [[1]](diffhunk://#diff-e8d59d232a103d745395193665c31c153d84d50db8e0ec82737b01e7dcbda865R27-R64) [[2]](diffhunk://#diff-e8d59d232a103d745395193665c31c153d84d50db8e0ec82737b01e7dcbda865R92-R122)

### Test Coverage:

* [`tests/metadata/test_definitions.py`](diffhunk://#diff-76584bee9566289f3d14a71b07706145003099f4382d2a0aaa3519896412399aR1): Updated tests to include `datetime` import and validate `DataType` mappings for `DATE` and `TIMESTAMP`. [[1]](diffhunk://#diff-76584bee9566289f3d14a71b07706145003099f4382d2a0aaa3519896412399aR1) [[2]](diffhunk://#diff-76584bee9566289f3d14a71b07706145003099f4382d2a0aaa3519896412399aL28-R38)
* [`tests/metadata/test_entities.py`](diffhunk://#diff-d7045374535e6171cede2c26a8d9228edf5d9e7167cdf819aa30c06048f8a940R1-R3): Added tests for `__setattr__` and `__repr__` methods in the `BaseEntity` class to ensure proper type handling and representation. [[1]](diffhunk://#diff-d7045374535e6171cede2c26a8d9228edf5d9e7167cdf819aa30c06048f8a940R1-R3) [[2]](diffhunk://#diff-d7045374535e6171cede2c26a8d9228edf5d9e7167cdf819aa30c06048f8a940R13-R40)

### Configuration:

* [`.coveragerc`](diffhunk://#diff-834e9b406d74791ffbafaeec9cc894082cea9739bc347b6ff9f72312c92ebc79R6): Updated coverage configuration to include files in `bam_masterdata/openbis/`.

### DIFF
--- a/bam_masterdata/metadata/entities.py
+++ b/bam_masterdata/metadata/entities.py
@@ -39,6 +39,7 @@ class BaseEntity(BaseModel):
             return
 
         if key in self._property_metadata:
+            # TODO add CONTROLLEDVOCABULARY and OBJECT cases
             expected_type = self._property_metadata[key].data_type.pytype
             if expected_type and not isinstance(value, expected_type):
                 raise TypeError(
--- a/bam_masterdata/metadata/entities.py
+++ b/bam_masterdata/metadata/entities.py
@@ -24,6 +24,43 @@ class BaseEntity(BaseModel):
     adding new methods that are useful for interfacing with openBIS.
     """
 
+    def __init__(self, **kwargs):
+        super().__init__()
+
+        # We store the `_property_metadata` during instantiation of the class
+        self._property_metadata = self.set_property_metadata()
+
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def __setattr__(self, key, value):
+        if key == "_property_metadata":
+            super().__setattr__(key, value)
+            return
+
+        if key in self._property_metadata:
+            expected_type = self._property_metadata[key].data_type.pytype
+            if expected_type and not isinstance(value, expected_type):
+                raise TypeError(
+                    f"Invalid type for '{key}': Expected {expected_type.__name__}, got {type(value).__name__}"
+                )
+
+        object.__setattr__(self, key, value)
+
+    def __repr__(self):
+        # Filter for attributes that are `PropertyTypeAssignment` and set to a finite value
+        fields = []
+        for key, metadata in self._property_metadata.items():
+            if isinstance(metadata, PropertyTypeAssignment):
+                value = getattr(self, key, None)
+                # Only include set attributes
+                if value is not None and not isinstance(value, PropertyTypeAssignment):
+                    fields.append(f"{key}={repr(value)}")
+
+        # Format the output
+        class_name = self.__class__.__name__
+        return f"{class_name}({', '.join(fields)})"
+
     @property
     def cls_name(self) -> str:
         """
@@ -50,6 +87,37 @@ class BaseEntity(BaseModel):
             )
         ]
         return [getattr(self, attr_name) for attr_name in base_attrs]
+
+    def set_property_metadata(self) -> dict:
+        """
+        Dictionary containing the metadata of the properties assigned to the entity type.
+
+        Returns:
+            dict: A dictionary containing the keys of the `PropertyTypeAssignment` attribute names and the
+            values of the definitions of `PropertyTypeAssignment`. Example:
+            {
+                "name": PropertyTypeAssignment(
+                    code="$NAME",
+                    data_type=VARCHAR,
+                    mandatory=True,
+                    property_label="Name"
+                ),
+                "age": PropertyTypeAssignment(
+                    code="AGE",
+                    data_type=INTEGER,
+                    mandatory=False,
+                    property_label="Age"
+                ),
+            }
+        """
+        cls_attrs = self.__class__.__dict__
+
+        # Store property metadata at class level
+        prop_meta_dict: dict = {}
+        for attr_name, attr_value in cls_attrs.items():
+            if isinstance(attr_value, PropertyTypeAssignment):
+                prop_meta_dict[attr_name] = attr_value
+        return prop_meta_dict
 
     def model_to_json(self, indent: Optional[int] = None) -> str:
         """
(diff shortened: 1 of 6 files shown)


JosePizarro3:
@carlosmada22 would you mind checking? You can test it works by setting properties like:
```python
my_object = Instrument(
    name="Testname",
)
```

This approach is similar to PyBIS when populating attributes, but I haven't implemented yet passing a JSON so it directly populates the properties.

bam_masterdata/metadata/entities.py
```diff
@@ -51,6 +89,37 @@ def _base_attrs(self) -> list:
         ]
         return [getattr(self, attr_name) for attr_name in base_attrs]
 
+    def set_property_metadata(self) -> dict:
```
carlosmada22, line 92: Since this function is returning the dictionary with the metadata of the entity, should it not be called `get_property_metadata(self) -> dict:` ?

tests/metadata/test_definitions.py
```diff
@@ -25,14 +26,16 @@ class TestDataType:
                 DataType.CONTROLLEDVOCABULARY,
```
carlosmada22, line 26: What type should be CONTROLLEDVOCABULARY?

carlosmada22 (approved):
Looks good! Very useful functions to have, and almost all the Data Types covered

carlosmada22:
> @carlosmada22 would you mind checking? You can test it works by setting properties like:
> 
> ```python
> my_object = Instrument(
>     name="Testname",
> )
> ```
> 
> This approach is similar to PyBIS when populating attributes, but I haven't implemented yet passing a JSON so it directly populates the properties.

Works as expected, you can set all the attributes and access them correctly:

```
my_object = Instrument(
    name="Testname",
    description="Testdescription",
    generated_code_prefix="TEST",
    auto_generated_codes=True,
)
```

But printing just the entity:
`print(my_object)`

Just the properties are printed. I am not sure if this is the correct behaviour.

bam_masterdata/metadata/entities.py
JosePizarro3: Yeah, you are right 👍🏻

tests/metadata/test_definitions.py
JosePizarro3: I think it should be an enumeration with the options being the `terms` assign to a Vocabulary.

I can leave it open as a string, as well.

JosePizarro3:
> Works as expected, you can set all the attributes and access them correctly:
> 
> ```
> my_object = Instrument(
>     name="Testname",
>     description="Testdescription",
>     generated_code_prefix="TEST",
>     auto_generated_codes=True,
> )
> ```
> 
> But printing just the entity: `print(my_object)`
> 
> Just the properties are printed. I am not sure if this is the correct behaviour.


Ok, so I should explain a bit more how this works.

For each entity type (I will call them "object type" from now on), we have 2 distinct attributes:
1. Definitions (grouped `defs`): these are immutable, i.e., the will not change when you fill in the ELN.
2. Assigned properties: these are mutable, i.e., when filling the ELN they change their value.

So my idea would be that, some user can actually instantiate the classes and assign properties to values. Something like:
```python
class MyExample(ObjectType):
    defs = ObjectTypeDefinitions(
        code="MY_EXAMPLE",
        description="An example of an object type with Definitions and Assigned Properties.",
        generated_code_prefix="MYEX",
        auto_generated_codes=True,
    )

    prop_1 = PropertyTypeAssignment(data_type=VARCHAR, ...)
    prop_2 = PropertyTypeAssignment(data_type=FLOAT, ...)
    prop_3 = PropertyTypeAssignment(data_type=TIMESTAMP, ...)


example = MyExample(
    prop_1="Hello",
    prop_2=2.3,
    prop_3=timestamp-whatever,
)
```

The motivation for this would be:
- Users can start creating scripts to populate the ELN using bam-masterdata for all of this. Something like pyBIS, but I think a bit more elegant, maybe?
- They could directly do `example.to_json()` and print out the dictionary in json format (or whatever other format if we implement this kind of methods). I think in pyBIS this is not so obvious how to do it, right?

I wonder whether all of this is worth it, whether we should do this via pyBIS instead, or how to handle it.

JosePizarro3:
Ok, I decided to add also some methods for `BaseEntity`, and their testing. These are: `to_json`, `to_dict`, `to_hdf5`. Hopefully now it is more clear, and it paves the way for when users want to print out these as JSON or HDF5 files.

I added the `h5py` dependency as well to handle HDF5 files

PR #86, opened 2025-02-19
82 convert current data model to json format

carlosmada22:
Description
This PR introduces functionality to convert the current Data Model into a structured JSON format, ensuring that each entity type (objects, collections, datasets, properties, vocabularies) is exported into its respective JSON file.

Key Changes

- Refactored DataModelLoader to generate a structured JSON representation of the Data Model.
- Handled special cases for property_types.py, ensuring properties are correctly extracted and stored separately.
- Each entity type is now saved as an full JSON file under checker/tmp/datamodel/ just to geeral checks, but everything is saved as a dict variable in the MasterdataChecker class, to be used later in the checks.
- Implemented a centralized logging mechanism so that both MasterdataChecker and DataModelLoader share the same logger.

Implementation Details
Entities are extracted dynamically from Python modules using inspect.
Properties are processed separately (like in the previous CLI implementation) and stored in the property_types section.
All entities saved as a variable in the parent Checker class to be used during checks.

### DIFF
--- a/bam_masterdata/checker/checker.py
+++ b/bam_masterdata/checker/checker.py
@@ -47,9 +47,9 @@ class MasterdataChecker:
         """
         Load new entities from various sources (Python classes, Excel, etc.).
         """
-        self.logger.info(f"Loading new entities from: {source} (Type: {source_type})")
-        loader = SourceLoader(source, source_type)
-        self.new_entities = loader.load()
+        # self.logger.info(f"Loading new entities from: {source} (Type: {source_type})")
+        # loader = SourceLoader(source, source_type)
+        # self.new_entities = loader.load()
 
     def validate(self, mode: str = "all") -> dict:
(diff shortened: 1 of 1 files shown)


JosePizarro3:
I didn't go into the details of the code, but from the description and the datamodel_loader.py, it seems to me there is some duplicated logic with the `bam_masterdata/cli/entities_to_json.py` module and the function there. I think both things could be combined, i.e., a single function which stores the Python classes in a single dictionary; then, `entities_to_json.py` takes care of separating them in different json files, while your logic here can reuse that function.

JosePizarro3:
Ok, #94 handles this.

Now, there is a module `cli/entities_dict` which loads what we discussed of a single dictionary per entity type OR a full dictionary with all entity types.

@carlosmada22 you can close this 🙂

PR #100, opened 2025-02-28
83 convert incoming masterdata to be checked to json format

carlosmada22:
### **Summary**
This pull request introduces significant improvements to the SourceLoader class by refining how entities are loaded from different source types (Python or Excel) and ensuring that the output format is correctly structured in the same way independently of the source type, matching the structure loaded for the current DataModel. The update includes better handling of enums (returned in the `EntitiesDict.single_json` function, in the `DataType `fields), improved logging, and additional test coverage.

### **Key Changes**
**Enhanced Source Detection:**

- Now detects if the source_path is either:
- A single .py file.
- A directory containing .py files.
- A single .xlsx file.
- Logs warnings for unsupported file types.

**Improved Data Conversion:**

_Python Source Handling:_

- Calls EntitiesDict.single_json() to load Python-based entity definitions.
- Converts enums (e.g., <DataType.VARCHAR: 'VARCHAR'> → 'VARCHAR') using the new convert_enums() function.

_Excel Source Handling:_

- Calls entities_to_json() to transform extracted entities into the desired JSON format.
- Ensures properties are stored as lists, with metadata stored separately under defs.

**Better Formatting for Output JSON:**

- Properties are now correctly structured as lists.
- Entity metadata (defs) is placed after properties for readability.
- Uses format_json_id() to ensure consistent PascalCase formatting for IDs.
- Converts empty strings ("") to None for fields like validation_script and vocabulary_code.

**Logging Improvements:**

- Logs the detected source type at the beginning.
- Warnings are displayed if an unsupported file type is provided.
- Comprehensive Testing:

**Added test cases to verify:**

- Correct source type detection.
- Proper logging of source type.
- Transformation of entities from both Python and Excel sources.
- Handling of enums using convert_enums().
- Overall correctness of the output format.

**Impact**

- This PR ensures that SourceLoader produces consistent, well-formatted JSON outputs regardless of the input format, for later comparisons and checks.
- The additional logging and validation steps improve debugging and usability.
- The updated test suite enhances confidence in correctness and stability.


carlosmada22:
> I can take a look.
> 
> Are the changes properly reflected with respect to main? I think the tracking is not working and there are more changes than what they should (e.g., the `checker.py` is here showing changes that already exists in main).
> 
> Do you mind checking this and see whether something is wrong with this and how to fix it if so? 🙂 maybe it is just about rebasing

Ok I found the problem, when merged the issue 101 (https://github.com/BAMresearch/bam-masterdata/pull/101), the branch that I have created for that issue, was generated from this branch and not from the main one, I must have clicked wrong when creating it, so when I merged that one, I also have merged the commits from this one in main.

JosePizarro3:
> > I can take a look.
> > Are the changes properly reflected with respect to main? I think the tracking is not working and there are more changes than what they should (e.g., the `checker.py` is here showing changes that already exists in main).
> > Do you mind checking this and see whether something is wrong with this and how to fix it if so? 🙂 maybe it is just about rebasing
> 
> Ok I found the problem, when merged the issue 101 (#101), the branch that I have created for that issue, was generated from this branch and not from the main one, I must have clicked wrong when creating it, so when I merged that one, I also have merged the commits from this one in main.

🤔 then, do you mind fixing the issue? Now there are different changes in main and here which are probably shared. So I think the solution is going to #101 and check the last commit (maybe here in Github), while checking these branch changes, side by side, and see whether there are inconsistencies and fix them

JosePizarro3:
I see! It is still weird that Github cannot properly track changes in the "Files changed" menu

Sadly, I'd like to review it but I trust on you. Maybe later I can directly do small fixes here and there.

carlosmada22:
> I see! It is still weird that Github cannot properly track changes in the "Files changed" menu
> 
> Sadly, I'd like to review it but I trust on you. Maybe later I can directly do small fixes here and there.

Yeah I know... Sorry for that, but is everything related with the SourceLoader and the conversions to the JSON format dicts, so we can check it later when you want. Also I found weird that the branch was generated from this one except from the main one, which is the one that I always use... but anyway, we have registered all the changes in the commits at least.

PR #102, opened 2025-03-04
Added new headers to excel_to_entities

carlosmada22:
Added new headers to excel_to_entities not present before:

For properties assigned:

- Metadata	
- Dynamic script
- Unique	
- Internal assignment

For dataset types:

- Main dataset pattern	
- Main dataset path


carlosmada22:
New fields still not present in the "fill_masterdata" function, this still needs to be defined. Just gathered and parsed in the excel function.

PR #107, opened 2025-03-18
85 perform the checks

carlosmada22:
### **Pull Request Description:**
This PR introduces several improvements to the Masterdata Checker, focusing on validation and comparison functionalities. The key updates include all the checks present in the discussion for the [current datamodel](https://github.com/BAMresearch/bam-masterdata/discussions/46)

- **Comparison Enhancements:**  
  - Added row location (`row_location`) in log messages for better debugging.  
  - Refined entity and property comparison logic to detect changes in attributes and flag inconsistencies.  
  - Improved error handling for properties that do not exist in openBIS.

- **CLI Improvements:**  
  - Enhanced error messages to better reflect issues in the current and incoming models.  
  - Streamlined logging and validation output to make debugging easier.

#### **Modified Files:**
- `checker.py`: Updated data model loading and validation flow.
- `masterdata_validator.py`: Refactored validation checks, improved section ordering logic, and added stricter property checks.
- `cli.py`: Enhanced CLI error messaging and validation results handling.

### DIFF
--- a/bam_masterdata/checker/checker.py
+++ b/bam_masterdata/checker/checker.py
@@ -8,66 +8,26 @@ from bam_masterdata.utils import load_validation_rules
 
 
 class MasterdataChecker:
-    def __init__(
-        self,
-        validation_rules_path: str = "bam_masterdata/checker/validation_rules",
-        datamodel_dir: str = "bam_masterdata/datamodel",
-    ):
+    def __init__(self):
         """
         Initialize the comparator with validation rules and set the datamodel directory.
 
         Args:
-            validation_rules_path (str): Path to the validation rules JSON file.
-            datamodel_dir (str, optional): Directory where the Python datamodel files are located.
-                                           Defaults to "bam_masterdata/datamodel".
+            mode (str): The validation mode ("self", "incoming", "validate", "compare", "all").
         """
-        self.validation_rules_path = (
-            validation_rules_path  # Store the base directory path
-        )
-        self.datamodel_dir = (
-            datamodel_dir  # Allows overriding the default datamodel directory
-        )
         self.current_model: dict = None
         self.new_entities: dict = None
         self.logger = logger
-        self.validation_rules: dict = {}  # Initialize empty validation rules
+        self.validation_rules: dict = {}
 
-    def _load_validation_rules(self, mode: str) -> dict:
-        """
-        Load validation rules dynamically based on the validation mode using utils.load_validation_rules.
-
-        Args:
-            mode (str): The validation mode ("self", "incoming", "validate", "compare", "all").
-
-        Returns:
-            dict: The validation rules loaded from the corresponding JSON file.
-        """
-        rule_file_map = {
-            "self": "validation_rules.json",
-            "incoming": "validation_rules.json",
-            "validate": "validation_rules.json",
-            "compare": "compare_validation_rules.json",
-            "all": "all_validation_rules.json",
-        }
-
-        rule_file = rule_file_map.get(mode)
-        if not rule_file:
-            raise ValueError(f"No rule file found for mode: {mode}")
-
-        rule_path = os.path.join(self.validation_rules_path, rule_file)
-
-        return load_validation_rules(
-            self.logger, rule_path
-        )  # Use the existing function
-
-    def load_current_model(self):
+    def load_current_model(self, datamodel_dir: str = "./bam_masterdata/datamodel/"):
         """
         Load and transform the current data model (Pydantic classes) into JSON.
 
         Uses the default datamodel directory unless overridden.
         """
-        self.logger.info(f"Loading current data model from: {self.datamodel_dir}")
-        entities_dict = EntitiesDict(python_path=self.datamodel_dir, logger=self.logger)
+        self.logger.info(f"Loading current data model from: {datamodel_dir}")
+        entities_dict = EntitiesDict(python_path=datamodel_dir, logger=self.logger)
         self.current_model = entities_dict.single_json()
 
     def load_new_entities(self, source: str):
@@ -78,7 +38,7 @@ class MasterdataChecker:
         loader = SourceLoader(source)
         self.new_entities = loader.load()
 
-    def validate(self, mode: str = "all") -> dict:
+    def check(self, mode: str = "all") -> dict:
         """
         Run validations.
 
@@ -115,8 +75,12 @@ class MasterdataChecker:
                 "New entities must be loaded before validation in 'incoming', 'validate', 'compare', or 'all' modes."
             )
 
-        # Load the appropriate validation rules based on the mode
-        self.validation_rules = self._load_validation_rules(mode)
+        # Load the validation rules
+        if (
+            mode in {"self", "incoming", "validate", "all"}
+            and self.validation_rules == {}
+        ):
+            self.validation_rules = load_validation_rules(self.logger)
 
         validator = MasterDataValidator(
             self.new_entities, self.current_model, self.validation_rules
--- a/bam_masterdata/checker/checker.py
+++ b/bam_masterdata/checker/checker.py
@@ -1,6 +1,6 @@
 import os
 
-from bam_masterdata.checker.masterdata_validator import MasterDataValidator
+from bam_masterdata.checker.masterdata_validator import MasterdataValidator
 from bam_masterdata.checker.source_loader import SourceLoader
 from bam_masterdata.logger import logger
 from bam_masterdata.metadata.entities_dict import EntitiesDict
@@ -80,7 +80,7 @@ class MasterdataChecker:
         ):
             self.validation_rules = load_validation_rules(self.logger)
 
-        validator = MasterDataValidator(
+        validator = MasterdataValidator(
             self.new_entities, self.current_model, self.validation_rules
         )
         return validator.validate(mode)
--- a/bam_masterdata/checker/checker.py
+++ b/bam_masterdata/checker/checker.py
@@ -8,12 +8,11 @@ from bam_masterdata.utils import load_validation_rules
 
 
 class MasterdataChecker:
+    VALID_MODES = {"self", "incoming", "validate", "compare", "all", "individual"}
+
     def __init__(self):
         """
         Initialize the comparator with validation rules and set the datamodel directory.
-
-        Args:
-            mode (str): The validation mode ("self", "incoming", "validate", "compare", "all").
         """
         self.current_model: dict = None
         self.new_entities: dict = None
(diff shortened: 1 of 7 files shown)


bam_masterdata/checker/checker.py
```diff
@@ -8,66 +8,26 @@
 
 
 class MasterdataChecker:
```
JosePizarro3, line 10: I would design this class a bit different: by looking at your `__init__`, it seems an older implementation evolved quite a lot.

My suggestion (pure class design, not really a strong argument from one to another implementation than only readability and style) would be to define:
```python
class MasterdataChecker:
    def __init__(self, mode: str = "self", **kwargs) -> None:
        """
        Initialize the comparator with validation rules and set the datamodel directory.

        Args:
            mode (str): The validation mode ("self", "incoming", "validate", "compare", "all").
        """
        self.mode = mode
        self.logger = kwargs.get("logger", logger)
        self.datamodel_dir = kwargs.get("datamodel_dir", "./bam_masterdata/datamodel")
```

and then I would define two methods returning dictionaries (note I am changing `load` for `get` in the function names): `def get_current_model(self)` and `def get_new_entities(self)`. In these methods, instead of assigning `self.current_model` and `self.new_entities`, you just return the respective dictionaries. Example:
```python
class MasterdataChecker:
    def __init__(self, mode: str = "self", **kwargs):
         ...

    def get_current_model(self):
        self.logger.info(f"Loading current data model from: {self.datamodel_dir}")
        entities_dict = EntitiesDict(python_path=self.datamodel_dir, logger=self.logger)
        return entities_dict.single_json()
```

and the same for `get_new_entities`.

Then, you just need to change `def check(self)` to accommodate these returns.

bam_masterdata/checker/checker.py
```diff
@@ -8,66 +8,26 @@
...
-    ):
+    def __init__(self):
         """
         Initialize the comparator with validation rules and set the datamodel directory.
 
         Args:
-            validation_rules_path (str): Path to the validation rules JSON file.
-            datamodel_dir (str, optional): Directory where the Python datamodel files are located.
-                                           Defaults to "bam_masterdata/datamodel".
+            mode (str): The validation mode ("self", "incoming", "validate", "compare", "all").
```
JosePizarro3, line 16: You can use a variable `VALID_MODES` as a list of these values

I saw below (in `check()` that you are using a set `{"self", "incoming", "validate", "compare", "all"}`. I would use a list.

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -1,7 +1,7 @@
-import json
 import re
 
 from bam_masterdata.logger import logger
+from bam_masterdata.metadata.definitions import DataType
 from bam_masterdata.utils import is_reduced_version
```
JosePizarro3, line 7: a small typo (I think): `class MasterdataValidator` (now `Data` is capitalize)

Furthermore, I have a similar suggestion as with the `MasterdataChecker` class. Instead of storing in the class mem the dictionaries, simply return then in the corresponding function.

I saw that you are actually returning, but still storing in `self` variables. I would simply reject this if these are not use somewhere else.

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -37,27 +37,48 @@ def validate(self, mode: str = "all") -> dict:
...
-        self.validation_results = {}
+        self.validation_results = {
+            "current_model": {},
+            "incoming_model": {},
+            "comparisons": {},
+        }
 
         if mode in ["self", "all", "validate"]:
             self.logger.info("Validating current model...")
             self._validate_model(self.current_model)
```
JosePizarro3, line 48: this is part of the suggestion i have: in this function you are anyways passing `self.current_model` as an input. If this is a class function and you stored `current_model` in a class variable, why do you need to pass it anyways?

Like I said before, I think this is just the classes evolving too much and the style got a bit mixed up, so no problem, I just noticed these :-)

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -143,13 +240,19 @@ def _validate_fields(
...
+            "terms_validation": "in 'terms'.",
         }.get(rule_type, ".")
 
         for field, value in data.items():
             rule = self.validation_rules.get(rule_type, {}).get(field)
 
+            log_message = (
+                f"Invalid '{value}' value found in the '{field}' field at line {row_location} "
+                f"in entity '{entity_name}' of '{entity_type}'"
+                + (f" {extra_location} " if extra_location else " ")
```
JosePizarro3, line 253: I would simply add the `if` separated from the long string to make this more readable

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -160,31 +263,35 @@ def _validate_fields(
...
-                        parent_entity,
-                        entity_type,
-                        entity_name,
-                        field,
-                        value,
-                        row_location,
-                        extra_location,
-                        "Invalid format.",
-                    )
+                    log_message = log_message + "Invalid format."
```
JosePizarro3, line 266: not a big fan of using `+` when defining strings: you lose a bit of control on spaces

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -160,31 +263,35 @@ def _validate_fields(
...
+                        log_message = (
+                            log_message
+                            + "Description should follow the schema: English Description + '//' + German Description. "
+                        )
+                    if "is_section" in rule:
+                        log_message = (
+                            log_message
+                            + " First letter of every word starts with capitalized lettter."
+                        )
+                    self._store_log_message(parent_entity, log_message, level="error")
```
JosePizarro3, line 277: these are definitely not errors, but warnings

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -105,6 +157,51 @@ def _validate_model(self, model: dict) -> dict:
                             entity_data,
                         )
 
+                # Check if "Additional Information" is followed only by "Additional Information" or "Comments"
```
JosePizarro3, line 160: this error is a bit... meh: who cares about the order...?

i would simply delete these lines

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -193,87 +300,272 @@ def _validate_fields(
...
-                        entity_type,
-                        entity_name,
-                        field,
-                        value,
-                        row_location,
-                        extra_location,
-                        "The generated code should be a part of the code.",
+                        log_message
+                        + "The generated code should be a part of the code.",
+                        level="error",
```
JosePizarro3, line 307: is this an error? I thought the `generated_code_prefix` could be a bit random

bam_masterdata/checker/masterdata_validator.py
```diff
@@ -193,87 +300,272 @@ def _validate_fields(
...
+                    log_message = (
+                        "Entity will not be imported in openBIS. "
+                        f"The entity {entity_code} at row {incoming_entity['defs'].get('row_location')} has the same properties defined as {other_entity_code}. "
+                        "Maybe they are representing the same entity?"
+                    )
+                    self._store_log_message(
+                        incoming_entity, log_message, level="warning"
+                    )
+
+    def _store_log_message(self, entity_ref, message, level="error"):
```
JosePizarro3, line 525: interesting...

shouldn't this be an external function that could be use for any class?

bam_masterdata/cli/cli.py
```diff
@@ -380,28 +381,72 @@ def export_to_rdf(force_delete, python_path, export_dir):
...
-    "--to",
-    "to_path",  # alias
+    "--datamodel-path",
+    "datamodel_path",  # alias
     type=click.Path(exists=True, dir_okay=True),
-    default=DATAMODEL_DIR,
-    help="""
-    The path to the directory containing the Python modules the the path defined in `from` will be checked with respect to.
-    """,
+    default="./bam_masterdata/datamodel/",
```
JosePizarro3, line 411: shouldn't be `DATAMODEL_DIR` the default?

JosePizarro3 (commented):
very cool, just some minor details

many thanks, it is actually looking awesome! I want to test this myself on the individual masterdata repos, and i am looking forward to integrate it in the github actions :-)

bam_masterdata/checker/checker.py
carlosmada22: If for you this looks better I can change it sure.

The only thing that I disagree here and I would like to keep is when assigning the `mode` as a class variable, since my idea is that you can specify the `mode` when running the `validate` function. but if you do it like that, you will need to specify the mode when creating the `MasterdataChecker` object, and if you want to run a different mode, you need to do this again. But if we leave it as it is now, you can just run the `validate` function specifying the desired `mode` when running it, which makes more sense to me.

carlosmada22: Another thing that I am noticing with this changes, is that every time that I run the checker, if I get rid of the class variables for the models (current and incoming) and the validation rules, each time that I run the checker, it will need to load over and over the same models each time, but as it is now, it will just load them once if it is not loaded, so I really think it is working fine as it is know, in terms of usability (I went over this topic and the mode a lot when designing it until I got the final idea which is the one coded here).

So in my opinion I think I would like to keep it as it is know. However if you think we should definitely go for the other way, I think we can discuss it more in deep to see how to change it 🙂

bam_masterdata/checker/checker.py
carlosmada22: Done!

bam_masterdata/checker/masterdata_validator.py
carlosmada22: The reason for that is because the same function is used for checking the incoming datamodel or the current one, because the format checks performed are the same ones, so I just specify it there for working with one model or the other one depending on the mode.

I think I am not getting right how you want to change it here, I think will be nice to have a little talk about this part so we can clarify our points

bam_masterdata/checker/masterdata_validator.py
carlosmada22: Done!

bam_masterdata/checker/masterdata_validator.py
carlosmada22: Done, changed for f-strings

bam_masterdata/checker/masterdata_validator.py
carlosmada22: It was specifically requested by Angela in the [checks discussion](https://github.com/BAMresearch/bam-masterdata/discussions/46)...:

> Additional Checks for the Masterdata Checker that I did not find in the list:
> 
> 1. Property $ANNOTATIONS_STATE is deprecated for openBIS 20.10.7.3.
> 
> 3. "**Section**" 
> - [x] Section names: First letter of every word starts with capitalized lettter
> - [x] Section "Additional Information" is followed by Section "Comments"
> - [x] Property "Notes" exists in Section  "Additional Information"
> - [x] Property "Comments" exists in Section  "Comments"

Not a big fan either, but I think this should be discussed with her before removing it, if she was really interested on this...

bam_masterdata/checker/masterdata_validator.py
carlosmada22: Done!

bam_masterdata/checker/masterdata_validator.py
carlosmada22: I can set it up as a warning, not a fixed thing for the `generated_code_prefix`

bam_masterdata/checker/masterdata_validator.py
carlosmada22: I can move it to `utils.py` if makes sense

carlosmada22: Done

bam_masterdata/cli/cli.py
carlosmada22: Correct, changed

PR #131, opened 2025-04-07
Add contributors to datamodel

JosePizarro3:
@CagtayFabry @bruehle do you mind checking the README, LICENSE and AUTHORS? 🙂  If you agree, I name you reviewers and if everything looks good a simple acceptance will work.

The whole list of contributors to the datamodel was extracted from talking with Angela.

I am also [excluding certain vocabularies](https://github.com/BAMresearch/bam-masterdata/pull/131/files#diff-72cf9acdc751bdb2072995fe55e20737f39e088aa8b74ee24e5593452fdb4eeeR159) to be exported to Python.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,7 +10,7 @@
 - Angela Ariza de Schellenberger
 - Ingo Bressler
 - Rukeia El-Athman
-- Cagtay Fabry
+- Çağtay Fabry
 - Tobias Fritsch
 - Ralf Herrmann
 - Zoltán Konthur
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,23 @@
+# This is the list of BAM Masterdata's significant contributors.
+#
+# This does not necessarily list everyone who has contributed code,
+# especially since many employees of one corporation may be contributing to the
+# definition of object types, property types, and vocabularies in the Masterdata
+# datamodel.
+# To see the full list of code contributors, see the revision history in
+# source control.
+
+- Angela Ariza de Schellenberger
+- Ingo Bressler
+- Rukeia El-Athman
+- Cagtay Fabry
+- Tobias Fritsch
+- Ralf Herrmann
+- Zoltán Konthur
+- Julius Kruse
+- Pavlina Kruzikova
+- Tarakeshwar Lakshmipathy
+- Julien Lecompagnon
+- Jan Lisec
+- Mathias Röllig
+- Bastian Rühle
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ The main datamodel developers are:
 - Angela Ariza de Schellenberger
 - Ingo Bressler
 - Rukeia El-Athman
-- Cagtay Fabry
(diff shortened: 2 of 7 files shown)


bruehle:
@JosePizarro3 , thanks, looks good to me. Perhaps you don't even have to exclude the masterdata definitions from the license now that everyone who contributed is acknowledged, and - from what I gathered from the emails - is also fine with having it publicly available? But we can of course also keep it this way for now, then we are on the safe side. And we can always change the license again later to something more permissive.

JosePizarro3:
@bruehle thanks!

> Perhaps you don't even have to exclude the masterdata definitions from the license now that everyone who contributed is acknowledged

I think I like the idea to keep it a bit more restrictive for now. The phrase is also a bit open to simply reference to this repo and the people here if the datamodel definitions are used (which is not such a bad thing) 🙂

README.md
```diff
@@ -146,11 +146,31 @@ INFO    -  [14:07:47] Watching paths for changes: 'docs', 'mkdocs.yml'
...
 | Carlos Madariaga    | [carlos.madariaga@bam.de](mailto:carlos.madariaga@bam.de)       |
 | Dr. Jose M. Pizarro | [jose.pizarro-blanco@bam.de](mailto:jose.pizarro-blanco@bam.de) |
+
+
+The main datamodel developers are:
+
+- Angela Ariza de Schellenberger
+- Ingo Bressler
+- Rukeia El-Athman
+- Cagtay Fabry
```
CagtayFabry, line 166:
```suggestion
- Çağtay Fabry
```

AUTHORS
```diff
@@ -0,0 +1,23 @@
...
+# especially since many employees of one corporation may be contributing to the
+# definition of object types, property types, and vocabularies in the Masterdata
+# datamodel.
+# To see the full list of code contributors, see the revision history in
+# source control.
+
+- Angela Ariza de Schellenberger
+- Ingo Bressler
+- Rukeia El-Athman
+- Cagtay Fabry
```
CagtayFabry, line 13:
```suggestion
- Çağtay Fabry
```

bam_masterdata/openbis/get_entities.py
```diff
@@ -153,6 +153,19 @@ def get_vocabulary_dict(self) -> dict:
...
+            if code in [
+                "BAM_FLOOR",
+                "BAM_HOUSE",
+                "BAM_LOCATION",
+                "BAM_LOCATION_COMPLETE",
+                "BAM_OE",
+                "BAM_ROOM",
+                "PERSON_STATUS",
+            ]:
+                continue
```
CagtayFabry, line 168: Thank you for including this change

CagtayFabry (changes requested):
Thanks for these changes, just some minor technical suggestions (if UTF encoding is feasable here)

JosePizarro3:
> Thanks for these changes, just some minor technical suggestions (if UTF encoding is feasable here)

Sure, thanks for the review!

PR #132, opened 2025-04-07
13 add push to openbis to update masterdata in openbis new version

carlosmada22:
**Push to openbis ready for all object types**

1. It first creates a copy of the Python files in a temporary folder, or converts the Excel file into Python files in the temporary folder (depending on the input type)
2. Checks that everything went well in the checker and that there are no errors or criticals (mode='individual').
3. If there are no problems, it goes file by file, and entity by entity, and runs the new "to_openbis()" method, inner on each class, that creates this new entity in the openbis instance defined in the environment variable (.env) (entity update is not allowed for the moment).
4. During the whole process, it indicates the user all the performed steps.

_EXTRAS_

- Now the order of the generated dictionaries is the same as the one present in the Python classes.
- Added "SAMPLE" as accepted data type (treated equally as "OBJECT", just to work with older entities defined as this one).


bam_masterdata/cli/cli.py
```diff
@@ -444,5 +447,105 @@ def checker(file_path, mode, datamodel_path):
...
+
+    # If there are no problems, push to openBIS
+    if all(
+        isinstance(value, dict) and all(not sub_value for sub_value in value.values())
+        for value in validation_results.values()
+    ):
+        click.echo("No problems found in the datamodel and incoming model.")
+        # Push to openBIS
+
+        url = environ("OPENBIS_URL")
```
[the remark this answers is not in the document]
JosePizarro3: nah, disregard this @carlosmada22

bam_masterdata/cli/cli.py
```diff
@@ -444,5 +447,105 @@ def checker(file_path, mode, datamodel_path):
...
+    else:
+        source_type = None
+        logger.warning(f"Unsupported source type for path: {file_path}")
+
+    # Handle source_type
+    tmp_dir = "./bam_masterdata/tmp"
+    os.makedirs(tmp_dir, exist_ok=True)
+
+    if source_type == "python":
+        # Copy all .py files to the tmp directory
```
JosePizarro3, line 485: not sure if i understand why to copy all Python files to the tmp directory

you can anyways do:
```python
if os.path.isdir(file_path):
    tmp_dir = file_path
else:
    tmp_dir = os.path.dirname(file_path)  # ! please check this os method is the correct one!!
```

tests/metadata/test_entities.py
```diff
@@ -97,16 +97,22 @@ class TestObjectType:
...
         prop_names = [prop.code for prop in object_type.properties]
-        assert prop_names == ["ALIAS", "$NAME"]
+        assert prop_names == ["$NAME", "ALIAS"]
 
         # 3 properties in this `ObjectType`
         object_type = generate_object_type_longer()
         assert len(object_type.properties) == 3
         prop_names = [prop.code for prop in object_type.properties]
-        assert prop_names == ["ALIAS", "$NAME", "SETTINGS"]
+        assert prop_names == ["SETTINGS", "$NAME", "ALIAS"]
```
JosePizarro3, line 115: ok this is a bit strange, i cannot reproduce

do you mind writing a `# TODO` to check this later on, most importantly in the specific method for the `entities.py`? something like:
```python
# TODO check if the order is properly assigned
```

JosePizarro3 (approved):
all good :-) looks really nice, thanks!

just couple of minor comments

bam_masterdata/cli/cli.py
carlosmada22: IT was mainly to keep the same behaviour no matter if the input was Excel or Python, always do the same steps.

PR #152, opened 2025-07-29
151 implementation of   the setattr method for objecttype object and controlledvocabulary

LucasZimm:
Added the __setattr__ logic for controlledvocabulary #TODO Objecttype

### DIFF
--- a/bam_masterdata/metadata/entities.py
+++ b/bam_masterdata/metadata/entities.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, no_type_check
@@ -6,6 +7,8 @@ import h5py
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 from rdflib import BNode, Literal
 from rdflib.namespace import DC, OWL, RDF, RDFS
+
+from bam_masterdata.utils import DATAMODEL_DIR, import_module, listdir_py_modules
 
 if TYPE_CHECKING:
     from pybis import Openbis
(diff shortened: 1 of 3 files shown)


bam_masterdata/metadata/entities.py
```diff
@@ -497,6 +500,50 @@ class ObjectType(BaseEntity):
...
+            for file in listdir_py_modules(DATAMODEL_DIR):
+                if "vocabulary_types.py" in file:
+                    vocab_path = file
+                    break
+
+            module = import_module(vocab_path)
+
+            for name, obj in inspect.getmembers(module, inspect.isclass):
+                if issubclass(obj, VocabularyType) and obj is not VocabularyType:
+                    if hasattr(obj, "defs") and obj.defs.code == vocabulary_code:
```
JosePizarro3, line 528: do you need this line? If in the previous `if` you are checking you have an instance of `VocabularyType`, I think you can simply skip this second if condition

bam_masterdata/metadata/entities.py
```diff
@@ -497,6 +500,50 @@ class ObjectType(BaseEntity):
...
+                        vocabulary_class = obj()
```
JosePizarro3, line 529: after this, you would need a `break` so it does not continue with all the for loop

bam_masterdata/metadata/entities.py
```diff
@@ -497,6 +500,50 @@ class ObjectType(BaseEntity):
...
+            # get the class of the vocabulary type
+            for file in listdir_py_modules(DATAMODEL_DIR):
+                if "vocabulary_types.py" in file:
+                    vocab_path = file
+                    break
+
+            module = import_module(vocab_path)
+
+            for name, obj in inspect.getmembers(module, inspect.isclass):
+                if issubclass(obj, VocabularyType) and obj is not VocabularyType:
```
JosePizarro3, line 527: can't you simplify this to `if name == vocabulary_code`?

bam_masterdata/metadata/entities.py
```diff
@@ -497,6 +500,50 @@ class ObjectType(BaseEntity):
...
+                    if hasattr(obj, "defs") and obj.defs.code == vocabulary_code:
+                        vocabulary_class = obj()
+
+            codes = []
```
JosePizarro3, line 531: I think this logic to get the `codes` of each term in a VocabularyType is already implemented by accessing the `terms` attribute. 

`terms` work for a VocabularyType in a similar fashion as `properties` do for ObjectType. So you could try:
```python
codes = [term.defs.code if term.defs for term in vocabulary_class.terms]
```

JosePizarro3 (approved):
Very cool, just some minor tweaks here and there, but all looking very nice, thanks! :-)

I also asked Github copilot to drop a review. I agree with its comments (at least the ones that are open or unresolved still).

Once you implement the comments, feel free to let me know and we merge after the pipeline passes!

PR #181, opened 2025-08-28
added space support

LucasZimm:
Change the run parser function to work with a space name instead of getting it from the login

### DIFF
--- a/bam_masterdata/cli/cli.py
+++ b/bam_masterdata/cli/cli.py
@@ -590,6 +590,7 @@ def run_parser(
     files_parser: dict[AbstractParser, list[str]] = {},
     project_name: str = "PROJECT",
     collection_name: str = "COLLECTION",
+    space_name: str = None,
 ) -> None:
     """
     Run the parsers on the specified files and collect the results.
@@ -614,7 +615,7 @@ def run_parser(
         return
 
     # Specify the space and project for the data
-    space = openbis.get_space(openbis.username)
+    space = openbis.get_space(space_name)
     if space is None:
         logger.error(
             f"Space {openbis.username} does not exist in openBIS. Please create it first."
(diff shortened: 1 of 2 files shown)


bam_masterdata/cli/cli.py
```diff
@@ -614,7 +615,7 @@ def run_parser(
         return
 
     # Specify the space and project for the data
-    space = openbis.get_space(openbis.username)
+    space = openbis.get_space(space_name)
```
JosePizarro3, line 618: You can set it to default being the `openbis.username`, like:
```python
    if space_name is None:
        space_name = openbis.username
    space = openbis.get_space(space_name)
```

bam_masterdata/cli/cli.py
```diff
@@ -727,7 +728,14 @@ def run_parser(
     required=True,
     help="OpenBIS collection name",
 )
-def parser(files_parser, project_name, collection_name):
+@click.option(
+    "--space_name",
```
JosePizarro3, line 732: I would keep hyphens instead of underscores: `--space-name` (note you have an underscore between "space" and "name")

bam_masterdata/cli/cli.py
```diff
@@ -727,7 +728,14 @@ def run_parser(
...
+    "space_name-name",  # alias
```
JosePizarro3, line 733: not sure what is the purpose of this alias, I would delete it

bam_masterdata/cli/cli.py
```diff
@@ -727,7 +728,14 @@ def run_parser(
...
+    type=str,
+    required=True,
```
JosePizarro3, line 735: if the `space_name` defaults to openbis.username, I would delete this required option

JosePizarro3 (commented):
Some minor stuff

bam_masterdata/cli/cli.py
JosePizarro3: Ok, can you patch it with `if`s? 
```python
if 'devel' in openbis.url:
    space_name = openbis.username
else:
    space_name = ...
```

carlosmada22 (approved):
Looks good you can merge it. Just perform some global testing to check that it works as expected.

PR #184, opened 2025-09-08
182 validation of datetime in masterdata and pybis

LucasZimm:
changed logic behind the _setattr_ function for object_types to validate datetime as strings for imports using pyBis

### DIFF
--- a/bam_masterdata/metadata/entities.py
+++ b/bam_masterdata/metadata/entities.py
@@ -546,6 +546,13 @@ class ObjectType(BaseEntity):
                     expected_type = meta[key].data_type.pytype
                     if expected_type is datetime.datetime and isinstance(value, str):
                         # Try validate datetime from string
+                        if isinstance(value, datetime.datetime):
+                            try:
+                                value = value.strftime("%Y-%m-%d %H:%M:%S")
+                            except ValueError:
+                                raise ValueError(
+                                    f"Invalid datetime format for '{key}': Expected ISO format string, got '{value}'"
+                                )
                         try:
                             datetime.datetime.fromisoformat(value)
                             expected_type = str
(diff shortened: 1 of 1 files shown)


bam_masterdata/metadata/entities.py
```diff
@@ -543,6 +544,22 @@ def __setattr__(self, key, value):
                 if key in meta:
                     # Typcheck
                     expected_type = meta[key].data_type.pytype
+                    if expected_type is datetime.datetime and isinstance(value, str):
+                        # Try validate datetime from string
+                        if isinstance(value, datetime.datetime):
```
carlosmada22, line 549: This part looks unreachable. If `value` is `str` (outer check), then it will never be a `datetime` (inner check). 

I think you should do it in separate checks (example of how this could be done):
```
if expected_type is datetime.datetime:
    if isinstance(value, datetime.datetime):
        pass  # keep as-is
    elif isinstance(value, str):
        try:
            value = datetime.datetime.fromisoformat(value)
        except ValueError:
            raise ValueError(
                f"Invalid datetime format for '{key}': Expected ISO format string, got '{value}'"
            )
    else:
        raise TypeError(
            f"Invalid type for '{key}': Expected datetime or ISO format string, got {type(value).__name__}"
        )
```

carlosmada22 (commented):
Check the loop part, if everything works like that, then we can keep it as a quick-fix

bam_masterdata/metadata/entities.py
LucasZimm: Yes, the loop is unreachabe i will fix that.

With the if isinstance(value, datetime.datetime): 
the value should still be changed to a string so it can be uploaded with pyBis.

PR #207, opened 2025-10-20
fix regression: cli/excel_to_entities.py was deleted by f46b10f80

ibresslerBAM:


JosePizarro3:
hi @ibresslerBAM, is this and #206 intended to be here or in your fork? 🙂

PR #209, opened 2025-10-21
allow Openbis session in ologin

CagtayFabry:
Currently, the library uses the `ologin` function open Openbis sessions in various places. `ologin` is dependent on a specific environmental variables.

To allow more flexible use of the library, this small patch enables the user to pass through existing Openbis sessions that might have been opened in a different context.

### DIFF
--- a/bam_masterdata/openbis/login.py
+++ b/bam_masterdata/openbis/login.py
@@ -3,16 +3,23 @@ from pybis import Openbis
 
 
 # Connect to openBIS
-def ologin(url: str = "") -> Openbis:
+def ologin(url: str | Openbis = "") -> Openbis:
     """
     Connect to openBIS using the credentials stored in the environment variables.
 
+    If an existing Openbis session is provided, the session is returned.
+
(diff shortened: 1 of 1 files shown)


CagtayFabry:
@JosePizarro3 I tried to keep the function "as is" for compatibility.

There are a few places of lines similar to this
```py
class OpenbisEntities:
    def __init__(self, url: str = ""):
        self.openbis = ologin(url=url)
```

I think to keep compatibility high it would make sense to move away from this pattern into something more general where sessions can be provided, e.g.

```py
class OpenbisEntities:
    def __init__(self, openbis: Openbis | str = None):
        self.openbis = ologin(url=openbis)
```

The default behaviour could replicate the current use of environmental variables.

Of course we could stick with the `url` argument name but I think this might be confusion later.

Let me know what you think

CagtayFabry:
I just saw this is till open.

I don't have the necessary rights to merge this myself .. :wink: @JosePizarro3

PR #230, opened 2025-11-20
Delete MeasurementsCollection, obsolete

ibresslerBAM:
I created the MeasurementsCollection during rollout phase 3 as a container for a set of measurements acquired in a single session (time/place) share the same set of basic metadata. But since measurements objects could be moved around freely I switched to parent-child linking to a common MeasurementSession object instead. Consequently, we do not use MeasurementsCollection and I discourage anybody else from doing so - it's a leftover


ibresslerBAM:
Thanks for clarification. It's fine for me. This PR is just for clean up, feel free to remove that type (by merging this).

PR #235, opened 2025-11-25
234 add threading to run parser function

### LINKED ISSUE #234: Add Threading to run_parser Function
The current implementation of the run_parser function runs synchronously and blocks the main process, especially when parsing large or multiple files. This can lead to performance bottlenecks and long wait times.

Adding threading would allow the parser to run independently of the main process and enable timeout logic, so threads can be safely terminated after a specified timeout_time.

LucasZimm:
Simple thread function for run_parser

Closes #234

### DIFF
--- a/bam_masterdata/cli/run_parser.py
+++ b/bam_masterdata/cli/run_parser.py
@@ -288,8 +288,8 @@ class RunParser:
             timeout (int, optional): Time in seconds for the parser to complete before timing out. Defaults to 300.
             max_workers (int, optional):  Number of threads to use. Defaults to 1.
         """
+        start = time.perf_counter()
         try:
-            start = time.perf_counter()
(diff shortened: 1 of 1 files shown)


bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +265,41 @@ def run_parser(
...
+def run_parser_withthreading(
+    openbis=None,
+    space_name: str = "",
+    project_name: str = "PROJECT",
+    collection_name: str = "",
+    files_parser: dict = {},
+    collection_type: str = "COLLECTION",
+    timeout: int = 300,
+    max_workers: int = 1,
+) -> None:
```
JosePizarro3: maybe we could return the time it took to the multi thread executor of run_parser() instead of nothing?

PR #236, opened 2025-12-02
add MouseMeasurement type for integrating MOUSE instrument metadata

ibresslerBAM:
Proposing a specialisation of EXPERIMENTAL_STEP to store basic metadata for a set of measurement on each sample with our MOUSE instrument (fyi: https://lookingatnothing.com/index.php/archives/4053 and here https://lookingatnothing.com/index.php/archives/4353)
Tested and scripts ready for upload ;)

### DIFF
--- a/bam_masterdata/datamodel/object_types.py
+++ b/bam_masterdata/datamodel/object_types.py
@@ -12698,7 +12698,7 @@ class Crystal(MatSimStructure):
 #   EXPERIMENTAL_GOALS, SPREADSHEET, REFERENCE, PUBLICATION, COMMENTS
 class MouseMeasurement(SaxsMeasurement):
     defs = ObjectTypeDef(
-        code="EXPERIMENTAL_STEP.MOUSE_MEASUREMENT",
+        code="EXPERIMENTAL_STEP.SAXS_MEASUREMENT.MOUSE_MEASUREMENT",
         description="""Metadata of SAXS measurements of sample at MOUSE // Metadaten der SAXS-Messungen einer Probe mit MOUSE""",
         generated_code_prefix="EXP.MOME_",
     )


bam_masterdata/datamodel/object_types.py
```diff
@@ -12694,6 +12694,104 @@ class Crystal(MatSimStructure):
     )
 
 
+class MouseMeasurement(ExperimentalStep):
```
JosePizarro3: I think this should be instead a sub-type of `SaxsMeasurement` right?

bam_masterdata/datamodel/object_types.py
```diff
@@ -12694,6 +12694,104 @@ class Crystal(MatSimStructure):
...
+        code="EXPOSURE_TIME_IN_SECONDS",
+        data_type="REAL",
+        property_label="Exposure time [s]",
+        description="""Exposure time in seconds//Belichtungszeit in Sekunden""",
+        mandatory=True,
+        show_in_edit_views=False,
+        section="Experiment Details",
+    )
+
+    responsible_person = PropertyTypeAssignment(
```
JosePizarro3: 🤔 can you elaborate about this property? 

Are you thinking on a situation in which you need to transfer some data from somewhere else (in this case, MOUSE), and then you want to assign the person generating such data in this property?

Is that a "Responsible Person"?

bam_masterdata/datamodel/object_types.py
```diff
@@ -12694,6 +12694,104 @@ class Crystal(MatSimStructure):
...
+        code="RESPONSIBLE_PERSON",
+        data_type="OBJECT",
+        property_label="Responsible person",
+        description="""Responsible person//Verantwortliche Person""",
+        mandatory=False,
+        show_in_edit_views=False,
+        section="Experiment Details",
+    )
+
+    sample_location = PropertyTypeAssignment(
```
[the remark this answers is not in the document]
ibresslerBAM: We where looking for a property to reuse and sample_location is of suitable type VARCHAR. The actual description would be 'the sample position ID in the sample holder', some kind of spatial/orientational configuration in the setup.

bam_masterdata/datamodel/object_types.py
JosePizarro3: You can set `show_in_edit_views=False,` for any PropertyTypeAssignment you want to hide. Now that I check this, I have the feeling that the nicer implementation should be: show_in_edit_views is default to True, and then only when marked to False is ignored.

Now, regarding `frame_count`, this is indeed a problem, but for what you say, it seems that there is: one base SaxsMeasurement class, with 2 specific instruments doing this. Translated to classes, this means that you should have another 2 classes, one for each instrument. For now, you can leave it as it is, but for the future we have to be careful with these definitions 🙂

bam_masterdata/datamodel/object_types.py
```diff
@@ -12694,6 +12694,104 @@ class Crystal(MatSimStructure):
     )
 
 
+class MouseMeasurement(ExperimentalStep):
+    defs = ObjectTypeDef(
+        code="EXPERIMENTAL_STEP.MOUSE_MEASUREMENT",
+        description="""Metadata of SAXS measurements of sample at MOUSE // Metadaten der SAXS-Messungen einer Probe mit MOUSE""",
+        generated_code_prefix="EXP.MM_",
```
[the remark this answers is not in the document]
JosePizarro3: Yes, MOME_ looks better. I was refering to MM as being too unspecific

bam_masterdata/datamodel/object_types.py
```diff
@@ -12694,6 +12694,80 @@ class Crystal(MatSimStructure):
...
+        object_code="PERSON.BAM",
+        property_label="Responsible person",
+        description="""Responsible person//Verantwortliche Person""",
+        mandatory=False,
+        show_in_edit_views=True,
+        section="General Information",
+    )
+
+    # not a vocabulary: different sample holders might get new names,
+    # or one-off sample holders might have a temporary definition.
```
JosePizarro3: deleting comment lines

```suggestion
```

ibresslerBAM:
Thanks for the suggestions! I accepted them all. (I did not know, descriptions can be that long :)

Pushing this to the main instance as soon as possible, we'd prefer ;) We're eager to start ingesting data :D

PR #240, opened 2025-12-03
New object type: particle size analysis by laser diffraction methods

JosePizarro3:
Implementing new object and vocabulary from @BjoernMie


CagtayFabry:
I have a question wrt the indication of units in `property_label`. @JosePizarro3 
The [Wiki ](https://datastore.bam.de/en/Previous_version_of_Wiki/datastore/stewards/masterdata-best-practices#representation-of-values-and-units-of-measurement-of-physical-quantities) still mentions the indication of units using a `[mm/s]` style syntax for `property_label`. Should we still follow this or is it obsolete?

I don't think it is covered in the Masterdata checker currently

PR #243, opened 2025-12-09
Added data type verification for new Sample type

carlosmada22:
Is this what you meant @JosePizarro3 ??


JosePizarro3:
Looks good, but shouldn't we change it in other places of the checker as well? I am not sure eh, so this might be enough.

PR #249, opened 2025-12-17
add weldment identifiers

CagtayFabry:
- adds `Joint Number`, `Layer Number`, `Bead Number` and `Weldment Number` identifier schema to all welding experiment objects

### DIFF
--- a/bam_masterdata/datamodel/object_types.py
+++ b/bam_masterdata/datamodel/object_types.py
@@ -8728,7 +8728,6 @@ class Weldment(ExperimentalStep):
         property_label="Joint Number",
         description="""Consecutive numbering of weld joints of a workpiece or component//Fortlaufende Numerierung von Schweißnähten an Werkstücken und Bauteilen""",
         mandatory=False,
-        show_in_edit_views=False,
         section="Identifiers",
     )
 
@@ -8738,7 +8737,6 @@ class Weldment(ExperimentalStep):
         property_label="Layer Number",
         description="""Consecutive numbering of weld layers for a parent joint//Fortlaufende Numerierung von Schweißlagen der übergeordneten Schweißnaht""",
         mandatory=False,
-        show_in_edit_views=False,
         section="Identifiers",
     )
 
@@ -8748,7 +8746,6 @@ class Weldment(ExperimentalStep):
         property_label="Bead Number",
         description="""Consecutive numbering of weld beads or tracks for a parent layer//Fortlaufende Numerierung von Schweißraupen der übergeordneten Schweißlage""",
         mandatory=False,
-        show_in_edit_views=False,
         section="Identifiers",
     )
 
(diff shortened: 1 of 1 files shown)


bam_masterdata/datamodel/object_types.py
```diff
@@ -8722,6 +8722,46 @@ class Weldment(ExperimentalStep):
         section="Identifiers",
     )
 
+    weld_joint_number = PropertyTypeAssignment(
+        code="EXPERIMENTAL_STEP.WELD_JOINT_NUMBER",
```
JosePizarro3: Any reason why the code starts with EXPERIMENTAL_STEP?

bam_masterdata/datamodel/object_types.py
```diff
@@ -8722,6 +8722,46 @@ class Weldment(ExperimentalStep):
...
+        data_type="INTEGER",
+        property_label="Joint Number",
+        description="""Consecutive numbering of weld joints of a workpiece or component//Fortlaufende Numerierung von Schweißnähten an Werkstücken und Bauteilen""",
+        mandatory=False,
+        show_in_edit_views=False,
```
JosePizarro3: I think this attr was not properly exported by pybis; this is related with the editability of the property in the ELN, so I am guesisng all of yours are always True, right?

bam_masterdata/datamodel/object_types.py
CagtayFabry: update codes to remove `EXPERIMENTAL_STEP`

bam_masterdata/datamodel/object_types.py
JosePizarro3: I will set a default value to True, so my recommendation is simply to delete this.

bam_masterdata/datamodel/object_types.py
```diff
@@ -8722,6 +8722,46 @@ class Weldment(ExperimentalStep):
...
+        section="Identifiers",
+    )
+
+    weld_layer_number = PropertyTypeAssignment(
+        code="WELD_LAYER_NUMBER",
+        data_type="INTEGER",
+        property_label="Layer Number",
+        description="""Consecutive numbering of weld layers for a parent joint//Fortlaufende Numerierung von Schweißlagen der übergeordneten Schweißnaht""",
+        mandatory=False,
+        show_in_edit_views=False,
```
JosePizarro3:
```suggestion
```

bam_masterdata/datamodel/object_types.py
```diff
@@ -8722,6 +8722,46 @@ class Weldment(ExperimentalStep):
...
+        section="Identifiers",
+    )
+
+    weld_bead_number = PropertyTypeAssignment(
+        code="WELD_BEAD_NUMBER",
+        data_type="INTEGER",
+        property_label="Bead Number",
+        description="""Consecutive numbering of weld beads or tracks for a parent layer//Fortlaufende Numerierung von Schweißraupen der übergeordneten Schweißlage""",
+        mandatory=False,
+        show_in_edit_views=False,
```
JosePizarro3:
```suggestion
```

JosePizarro3 (approved):
You can add all my suggestions to a "batch" (from the FILES tab), and commit directly without the need of doing it locally

PR #259, opened 2026-02-04
Redefine datamodel

### LINKED ISSUE #155: Redefine datamodel
After the TF initial effort, I am going to update the `datamodel`  definitions

JosePizarro3:


bam_masterdata/datamodel/v2/base.py
```diff
@@ -0,0 +1,455 @@
+from bam_masterdata.metadata.definitions import ObjectTypeDef, PropertyTypeAssignment
+from bam_masterdata.metadata.entities import ObjectType
+
+# A list of deprecated ObjectType codes that should not be used in new data entries. This can be used to
+# maintain backward compatibility while signaling to users that certain types are no longer recommended for use.
+DEPRECATED_OR_UNUSED = [
+    "RAW_MATERIAL.STEEL",
+    "RAW_MATERIAL.ALUMINIUM",
+]
```
CagtayFabry, line 9: Just leaving a note here for later

I wasn't aware these types are deprecated. We actually started using `RAW_MATERIAL.STEEL` extensively lately

JosePizarro3: Please, do not use these instances as objects in your ELN/Inventory, @CagtayFabry

This was a design choice that, honestly, was wrong, imho. Nothing bad with who implemented it, because a lot of Data Modelling projects fall into the same problem. In our case, with our current aim being MSE (and not a sub-topic inside it) the schema should be clean and only contain objects, and not instances of the same.

If instances were to be included at the schema level, then 1) the purpose separation of objects and instances would become impossible to handle (again, thinking on MSE as a whole, and not a subtopic within it), 2) the schema itself will explode with all the posibilities (I mean, steel is just one amongst millions of options for RAW_MATERIAL), breaking the SQL system behind it. That's why we have the ELN, to create instances of objects.

Do you think we should clarify this for everybody?

CagtayFabry: Thank you for the explanation @JosePizarro3 
I will probably need some more input for clarification, maybe it is just about wording/naming of these object types?

What we need to represent is individual physical steel parts (like bars, pipes, sheets), each with individual measurements, comments etc. . These will be combined into the welded components. I think the original object type names `RAW_MATERUAL" was based on a more colloquial use in German language, since we often order those parts as "raw material".

Maybe we should discuss how this could be represented better in the Data Store soon?
Should we instantiate individual material classes like steel alloys + treatments and link to those?

JosePizarro3: > I will probably need some more input for clarification, maybe it is just about wording/naming of these object types?

The Object Type (i.e. what we define in the schema/Masterdata) should remain something like `RAW_MATERIAL` (or simply `MATERIAL`). Then, in the ELN/Inventory, we instantiate actual items, e.g. a specific bar, pipe, or sheet, whose material is steel.

Right now, `STEEL` is itself modeled as an Object Type. The issue with that approach is that it does not scale. You can think of all the possible combinations of elements, concentrations, symmetries, treatments, etc. that define a material. If we start encoding those at the schema level, we would end up needing an unbounded number of Object Types.

Another way to see it: if we allow `RAW_MATERIAL.STEEL`, why not also `RAW_MATERIAL.STEEL.xFe(1-x)C`? If the answer is “because we do not want to represent infinite variations of x”, then the same argument already applies one level before for “steel”, “aluminium”, “iron oxides”, etc. These are still just specific points in a much larger space of possibilities.

> What we need to represent is individual physical steel parts (like bars, pipes, sheets), each with individual measurements, comments etc. . These will be combined into the welded components. I think the original object type names `RAW_MATERUAL" was based on a more colloquial use in German language, since we often order those parts as "raw material".
> 
> Maybe we should discuss how this could be represented better in the Data Store soon?
> Should we instantiate individual material classes like steel alloys + treatments and link to those?

This sounds like two separate concerns:

1. Material definition (composition): e.g. steel, alloy/grade, treatment, standard, etc.
2. Physical realization (geometry): e.g. bar, pipe, sheet, with dimensions, batch, etc.

A cleaner model could separate these:

- schema Object Type: MATERIAL / RAW_MATERIAL
- (optional) material specification/masterdata: describing composition, grade, treatment, etc.
- ELN/Inventory instance: the actual physical part (bar/pipe/sheet), which references its material (perhaps via an OBJECT data type)

From that perspective, geometry (bar, sheet, pipe, etc.) is orthogonal to composition. One could model geometry as its own concept (either as a property or as a separate object linked to the instance), which then references the material it is made of. Though there are multiple valid ways to structure this (e.g. geometry-first vs material-first), **but I believe my argument holds: avoid encoding specific material instances directly as schema Object Types.**

I agree it would be worth discussing this more systematically for the Data Store, since different interpretations at different ontologies/schemas for MSE seem to have been used so far. We also have the limitation of the platform.

CagtayFabry: > The Object Type (i.e. what we define in the schema/Masterdata) should remain something like `RAW_MATERIAL` (or simply `MATERIAL`). Then, in the ELN/Inventory, we instantiate actual items, e.g. a specific bar, pipe, or sheet, whose material is steel.

> A cleaner model could separate these:
> schema Object Type: MATERIAL / RAW_MATERIAL
(optional) material specification/masterdata: describing composition, grade, treatment, etc.
ELN/Inventory instance: the actual physical part (bar/pipe/sheet), which references its material (perhaps via an OBJECT data type)

Basically what you suggest is already what we are doing now. We instantiate the actual items with the properties indicating:
- material geometry (bar, pipe, sheet ...) and dimensions
- material class according to DIN EN 10027-2 (those are linked to entries in `/BAM_GLOBAL/MISC/RAW_MAT_CODES` because the entries are too big to store as controlled vocabulary, it is already 2661 entries long just for steel alloys...)
The material code covers the specific alloys that I think you were hinting at with `RAW_MATERIAL.STEEL.xFe(1-x)C`. We don't use 100% Fe and the class was also never intended to represent that, although I understand the confusion given the name (I didn't create the OBJECT_TYPE, just wanted to reuse something that already existed)
- additional treatment specifiers from controlled vocabulary `RAW_MAT_TREATMENT_STEEL`

Maybe a more fitting `OBJECT_TYPE` name could be `RAW_MATERIAL.STEEL_ALLOY` or simply `STEEL_PART` or similar to avoid the confusion with the very broad `MATERIALS` term ?
Or do you want to have only one  `MATERIAL` OBJECT_TYPE for all specimen at BAM?

Let's discuss this in person soon, I think it is easier to show in the Data Store directly :)

bam_masterdata/datamodel/v2/activities.py
```diff
@@ -0,0 +1,1497 @@
...
+        A Processing is an activity that alters the structure, composition, or form of a
+        material.
+        An experimental step describing a welding experiment
+        """,
+        iri="https://bam.de/masterdata/Weldment",
+        references=[],
+        generated_code_prefix="WELDM",
+        aliases=[],
+        previous_versions=["EXPERIMENTAL_STEP.WELDMENT"],
+    )
```
CagtayFabry, line 1086:
```suggestion
class Welding(Processing):
    defs = ObjectTypeDef(
        code="WELDING",
        description="""
        A Processing is an activity that alters the structure, composition, or form of a
        material.
        An experimental step describing a general welding experiment
        """,
        iri="https://bam.de/masterdata/Welding",
        references=[],
        generated_code_prefix="WELD",
        aliases=[],
        previous_versions=["EXPERIMENTAL_STEP.WELDMENT"],
    )
```

After the recent discussions, I think this terminology is more in line with the new `Activity` and `Processing` types

bam_masterdata/datamodel/v2/base.py
```diff
@@ -0,0 +1,452 @@
...
+        """,
+        iri="https://bam.de/masterdata/InformationObject",
+        references=[
+            "http://purl.obolibrary.org/obo/IAO_0000030",
+        ],
+        generated_code_prefix="INFOR_OBJEC",
+        aliases=[],
+        previous_versions=[],
+    )
+
```
aradermacher, line 367: Would it be possible to add a XML property e.g. as PREVIEW at that level.
Than we could use that object to describe input/output files which might be yaml/json and put the content as preview here allowing a simple query.
If there will be a json/yaml property in teh future we should change that then.

aradermacher (commented):
I have one comment

bam_masterdata/datamodel/v2/base.py
JosePizarro3: Ok, sounds good to me. It could be a property of all the object types, in fact.

bam_masterdata/datamodel/v2/base.py
```diff
@@ -0,0 +1,452 @@
...
+        description="""
+        A Test is an activity that subjects materials to specific conditions to evaluate
+        their performance, reliability, or compliance with certain standards.
+        """,
+        iri="https://bam.de/masterdata/Test",
+        generated_code_prefix="TEST",
+        aliases=[],
+        previous_versions=[],
+    )
+
```
aradermacher, line 331: I'm not an experimentalist, but maybe Test and Measurement could be summarized or like this
class Test(Activity), class Measurement(Test)

bam_masterdata/datamodel/v2/base.py
```diff
@@ -0,0 +1,452 @@
...
+        material.
+        """,
+        iri="https://bam.de/masterdata/Processing",
+        generated_code_prefix="PROCE",
+        aliases=[],
+        previous_versions=[],
+    )
+
+
+class Simulation(Activity):
```
aradermacher, line 291: As discussed: We would like to have one more abstract level before Simulation for a "calling a script".
I propose: 
class Computation(Activity): ... description="""A Computation is a virtual activity that uses code/software to transfer/process input data into output data"""

compute_env = PropertyTypeAssignment(
        code="COMPUTE_ENV",
        data_type="MULTIVARCHAR",
        property_label="Description of the code environment",
        description="""
        Description of the code environment e.g. conda file, package, language.
        """,
        mandatory=False,
        section="General",
    )

class Simulation(Computation): ...

maybe also class Analysis(Computation) depends on if analysis is also used for physical analysis

PR #276, opened 2026-05-11
add welding datamodel subfolder

CagtayFabry:
Trying to clean up a bit before proposing new changes.

## Changes
- add `welding` subfolder for welding experiment related datamodels (`object_types.py`, `vocabulary_types.py`)
- move existing object types and vocabularies (no changes)
- removed unused `StationLayout1` and `StationLayout3` object type definitions
- rename `StationLayout2` to `StationLayout`


CagtayFabry:
Running the checker I get errors like the following for the classes moved into the subfolder (even though I removed them from the base datamodel files):

```
[critical ] The entity EXPERIMENTAL_STEP.WELDMENT.GMAW_BASE already exists in `bam-masterdata`. Please, check your classes.
```

bam_masterdata/datamodel/welding/__init__.py
JosePizarro3, line 1: no need for an `__init__`, unless you want to import the classes from it (which I guess you don't)

bam_masterdata/datamodel/welding/vocabularies.py
```diff
@@ -0,0 +1,78 @@
...
+        code="WELDING_SURFACING_WELD",
+        label="surfacing weld",
+        description="""A weld applied to a surface, as opposed to making a joint, to obtain desired properties or dimensions.//Auftragsschweißung zur Strukturbildung oder Beschichtung""",
+    )
+
+    welding_tack_weld = VocabularyTerm(
+        code="WELDING_TACK_WELD",
+        label="tack weld",
+        description="""Used to hold the parts of a weldment in proper alignment are placed in grooves or fillet locations and are small enough to be consumed by the production weld.//Heftnaht zur Positionierung von Bauteilen""",
+    )
```
JosePizarro3, line 78: Mmm something is still off with the GH actions, because `pre-commit run --all-files` (in actions.yml) should have complained about a missing end line in all these files...

Do you mind running that command locally and directly committing+adding all changes? I can take a look into why forking and creating a PR has this mess. But due to some other bugs, I am starting to hate forking.

bam_masterdata/datamodel/welding/vocabularies.py
JosePizarro3, line 1: just nitpicking: should we call these files `vocabularies.py` instead of `vocabulary_types.py`? I mean, I do not care that much, so we can leave it as it is

JosePizarro3 (commented):
Looks good to me, and I like the idea of defining sub-folders as application-specific definitions.

Just be mindful of the actions; as I said, I think they are not being run still when opening a PR from a forked repo.

bam_masterdata/datamodel/welding/vocabularies.py
JosePizarro3: ah, or is it the github web ui that is not showing properly that there is indeed an empty line before the end of the file??

bam_masterdata/datamodel/welding/__init__.py
CagtayFabry: I included it here mostly to add the submodule docstring

I can remove it if you prefer

bam_masterdata/datamodel/welding/vocabularies.py
CagtayFabry: locally I have empty lines in the files, I think this is how Github displays it (which is somewhat weird, it gives you a warning icon if there is no empty line)

Technical explanation seems to be this: https://dan-niles.medium.com/why-github-prs-shows-that-red-circle-understanding-newlines-at-end-of-files-8a4869821694

I will make sure to run pre-commit as well locally

bam_masterdata/datamodel/welding/vocabularies.py
CagtayFabry: I like `vocabularies.py`, just followed what was described in the HowTo guide

I will change it here and update the guide in a different PR

CagtayFabry:
this should be ready now @JosePizarro3

JosePizarro3:
merge at will, thanks!

Issue #286, opened 2026-05-13
Add dataset attachment to objects instead of collections when parsing


JosePizarro3:
Useful if files are to be attached to objects instead of the collection where these are generated (e.g., when attaching data or plots)

JosePizarro3:
Adding something like this in `bam_masterdata.metadata.entities.py:ObjectType`:
```python
    @property
    def dataset(self):
        pass

    def add_dataset(self, file):
        pass
```

And modifying the corresponding run_parser() logic to only store in the attached_object and not in the collection

PR #291, opened 2026-05-18
Add Transaction support for parsing into OpenBis

### LINKED ISSUE #263: Transactions for parsing
@LucasZimm 

Right now, parsing happens linearly with eahc new Object Type attached to a Collection in `run_parser()`. We could measure if transactions (see [pybis documentation](https://pypi.org/project/pybis/)) is faster. For this, we would need:

- Create or adapt the function `run_parser()` for the logic of saving objects.
- Explore if we could do batches of transactions to avoid problems.
- If batched transactions are an option, make sure this does not break relationships (i.e., an object A is created in batch 1 while it has relationships with object B which is created in batch 2)

Besides that, please read information about why transactions are faster in SQL platforms.

And besides normal testing, it would be nice to measure how fast is transactions vs linear creation of entities. This could be added to the documentation (i.e., number of entities created vs time taken)

LucasZimm:
Closes #263 

Adds transactions to the parsing workflow.


bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+        )
+
+    try:
+        trans2.commit()
+        logger.info("Datasets and relationships committed successfully")
+    except Exception as e:
+        logger.error(f"Failed to commit datasets/relationships: {e}")
+
+
+def _parser_init(
```
JosePizarro3: maybe a matter of taste, but can you put these functions at the beginning of the file?

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+    openbis: Openbis | None = None,
+    space_name: str = "",
+    project_name: str = "PROJECT",
+    collection_name: str = "",
+    files_parser: dict[AbstractParser, list[str]] = {},
+    collection_type: str = "COLLECTION",
+):
```
JosePizarro3: add docstring """

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+def _load_object_props(
+    object_id,
+    object_instance,
+    openbis,
+    collection,
+    openbis_id_map,
+    collection_name,
+    space_name,
+    project_name,
+):
```
JosePizarro3: add docstring and typing

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+                )
+                continue
+        else:
+            # Not an OBJECT type, handle normally
+            obj_props[property_metadata.code.lower()] = value
+
+        return obj_props
+
+
+def make_unique_code(code, seen_codes):
```
JosePizarro3: docstring + typing of the inputs/outputs of this function

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+    collection, space, project, collection_openbis = _parser_init(
+        openbis=openbis,
+        space_name=space_name,
+        collection_name=collection_name,
+        project_name=project_name,
+        files_parser=files_parser,
+        collection_type=collection_type,
+    )
+
+    trans = openbis.new_transaction()
```
JosePizarro3:
```python
    obj_transaction = openbis.new_transaction()
```

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+        openbis_id_map[object_id] = identifier
+
+    try:
+        trans.commit()
+        logger.info("Transaction committed successfully")
+    except Exception as e:
+        logger.error(f"Failed to commit object transaction: {e}")
+        return
+
+    trans2 = openbis.new_transaction()
```
JosePizarro3:
```python
    rel_transaction = openbis.new_transaction()
```

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+
+    # ---- DATASETS IN TRANSACTION ----
+    for files in files_parser.values():
```
JosePizarro3, line 465: would it be possible to define a function here for when storing the datasets in the collection? something that can be extended to objects in #286

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+            logger.info(f"{identifier} will be CREATED in transaction")
+
+        openbis_id_map[object_id] = identifier
+
+    try:
+        trans.commit()
+        logger.info("Transaction committed successfully")
+    except Exception as e:
+        logger.error(f"Failed to commit object transaction: {e}")
+        return
```
JosePizarro3:
```suggestion
        return None
```

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+                f"Duplicate local code {original_code} → renamed to {unique_code}"
+            )
+
+        identifier = (
+            f"/{space_name}/{project_name}/{unique_code}"
+            if not collection_name
+            else f"/{space_name}/{project_name}/{collection_name}/{unique_code}"
+        )
+
+        try:
```
JosePizarro3, line 429: is this try/except block necessary? I would imagine that get_object() returns None if identifier does not coincide with any existing object

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+def run_parser_with_transactions(
+    openbis: Openbis | None = None,
+    space_name: str = "",
+    project_name: str = "PROJECT",
+    collection_name: str = "",
+    files_parser: dict[AbstractParser, list[str]] = {},
+    collection_type: str = "COLLECTION",
+) -> None:
+    """
+    Run the parsers on the specified files and collect the results.
```
JosePizarro3: be more specific about what this function does with respect to `run_parser()`

bam_masterdata/cli/run_parser.py
```diff
@@ -263,3 +143,317 @@ def run_parser(
...
+    login with save_token=True don't forget!!
```
JosePizarro3: you can delete this line (same for `run_parser()`

docs/assets/transaction/Transaction vs SingleUpload.png
JosePizarro3, line 1: change labels and title to english

Zeitmessung -> Times
Zeit (Sekunden) -> Time (s)
Anzahl der Objekte -> Number of Created Objects
Normal -> without transactions
Transaction -> with transactions

JosePizarro3 (approved):
very good, some minor cosmetic stuff

bam_masterdata/cli/run_parser.py
LucasZimm: It will return an Error:


    raise ValueError(f"no such sample found: {sample_ident}")
    ValueError: no such sample found: test

bam_masterdata/cli/run_parser.py
LucasZimm: I would try doing that in #286 branch after this pull request is done.

bam_masterdata/cli/run_parser.py
JosePizarro3: Ok, weird.

Anyway, we leave the try/except block then, thanks for the info.

JosePizarro3:
Ok, merge when you want, thanks!

PR #299, opened 2026-05-26
286 add dataset attachment to objects instead of collections when parsing

### LINKED ISSUE #286: Add dataset attachment to objects instead of collections when parsing

LucasZimm:
Entities:
- adds the datasets property for ObjectType
- can use object.add_dataset to add a filepath

Parser:
- added feature to parse datasets attached to objects

Later Todo:
- dataset saving is not yet transaction compatible 

Fixed:
- Fixed Loop for Object property generation (could only do one property)


bam_masterdata/metadata/entities.py
```diff
@@ -968,6 +982,14 @@ def _assign_property(prop, entity, openbis) -> None:
         entity.save()
         return entity
 
+    @property
```
JosePizarro3: you can delete this property (`self.datasets` is already defined in pydantic)

bam_masterdata/cli/run_parser.py
```diff
@@ -459,6 +459,25 @@ def run_parser_with_transactions(
         logger.error(f"Failed to commit object transaction: {e}")
         return None
 
+    # TODO if later transactions support datasets change it to transaction.
```
JosePizarro3: just cherrypicking: maybe add the date where you wrote this (e.g., `" TODO (May 2026) ..."`)?

JosePizarro3 (approved):
Lovely, thanks!

PR #302, opened 2026-05-28
Fix issue 292: amending for mass spectrometry specific objects, properties and controlled vocabularies

janlisec:
existing object types amended by new properties:
- INSTRUMENT.MASS.SPEC
- EXPERIMENTAL_STEP.MS_BATCH
new object types:
- INSTRUMENT.LC_SYSTEM
- INSTRUMENT.GC_SYSTEM
several new contr.vocab
!! I removed (outcommented) one existing property (MS_IONIZATION_MODE) to avoid redundancy with a novel property MS_ION_POLARITY -- please check if this causes inconsistencies. If so I would modify the PR and use the old property.


bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
-    ms_ionization_mode = PropertyTypeAssignment(
-        code="MS_IONIZATION_MODE",
-        data_type="VARCHAR",
-        property_label="Ionization mode",
-        description="""Ionization mode (pos/neg)//Ionisierung (pos/neg)""",
-        mandatory=False,
-        section="MS Information",
-    )
+    # changes
+    # I would remove this property, as it is redundant with MS_ION_POLARITY. However should it be in use already, I would keep this one and not establish MS_ION_POLARITY.
```
JosePizarro3: I will check this

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
+    #    description="""Ionization mode (pos/neg)//Ionisierung (pos/neg)""",
+    #    mandatory=False,
+    #    section="MS Information",
+    # )
 
     ms_hyphenation_method = PropertyTypeAssignment(
         code="MS_HYPHENATION_METHOD",
         data_type="VARCHAR",
         property_label="Hyphenation method",
         description="""Hyphenation (DI, LC, GC, CE)//Probeninjektion (DI, LC, GC, CE)""",
```
JosePizarro3: Are these the only options or can there be more?

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
+        property_label="Resolution",
+        description="""Approximate Resolution//Ungefähre Auflösung""",
+        mandatory=False,
+        section="MS Parameters",
+    )
+
+    ms_intensity_range = PropertyTypeAssignment(
+        code="MS_INTENSITY_RANGE",
+        data_type="VARCHAR",
+        property_label="Intensity range",
```
JosePizarro3: if this is a range, I would like to see something else than a VARCHAR, perhaps a min/max couple of REAL properties (and the units)

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
         mandatory=False,
-        section="MS Information",
+        section="MS Parameters",
+    )
+
+    ms_resolution = PropertyTypeAssignment(
+        code="MS_RESOLUTION",
+        data_type="REAL",
+        property_label="Resolution",
+        description="""Approximate Resolution//Ungefähre Auflösung""",
```
JosePizarro3: Do you mind elaborating? is this some sort of pixel or m/z?

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
+    ms_intensity_range = PropertyTypeAssignment(
+        code="MS_INTENSITY_RANGE",
+        data_type="VARCHAR",
+        property_label="Intensity range",
+        description="""Ion intensity range//Ionen-Intensitätsbereich""",
+        mandatory=False,
+        section="MS Parameters",
+    )
+
+    ms_scan_rate = PropertyTypeAssignment(
```
JosePizarro3: yeah, I am a bit confused by this and the `ms_resolution`

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
+        code="MS_SCAN_RATE",
+        data_type="VARCHAR",
```
JosePizarro3: is this a VARCHAR or a REAL?

bam_masterdata/datamodel/object_types.py
```diff
@@ -9146,22 +9146,80 @@ class MsBatch(ExperimentalStep):
...
+        vocabulary_code="MS_ION_POLARITY",
+        property_label="Polarity",
+        description="""Ionization Polarity//Polarität der Ionisierung""",
+        mandatory=False,
+        section="MS Parameters",
+    )
+
+    ms_massrange = PropertyTypeAssignment(
+        code="MS_MASSRANGE",
+        data_type="VARCHAR",
```
JosePizarro3, line 7421: VARCHAR or REAL composed as min/max+units?

bam_masterdata/datamodel/object_types.py
```diff
@@ -9281,6 +9339,93 @@ class MassSpec(Instrument):
         section="Technical Details",
     )
 
+    # changes
```
JosePizarro3: you can delete these inline comments

bam_masterdata/datamodel/object_types.py
```diff
@@ -9281,6 +9339,93 @@ class MassSpec(Instrument):
...
+        description="""Detector Type//Analysator""",
+        mandatory=False,
+        section="Technical Details",
+    )
+
+
+class LcSystem(Instrument):
+    defs = ObjectTypeDef(
+        code="INSTRUMENT.LC_SYSTEM",
+        description="""LC-System//LC-System""",
```
JosePizarro3: maybe be more descriptive in all definitions? ChatGPT can help you with this without overdoing it

bam_masterdata/datamodel/object_types.py
```diff
@@ -9281,6 +9339,93 @@ class MassSpec(Instrument):
...
+        data_type="CONTROLLEDVOCABULARY",
+        vocabulary_code="DETECTOR_TYPE",
+        property_label="Detector",
+        description="""Detector Type//Analysator""",
+        mandatory=False,
+        section="Technical Details",
+    )
+
+
+class GcSystem(Instrument):
```
JosePizarro3: Mmm, for now, `LcSystem` and `GcSystem` contain the same properties, so I'd argue that both can be simply defined as a single entity and then add something like `kind = ["LC", "GC"]` to differentiate between them.

It is even related with CHROMATOGRAPHY_TYPE in the `MassSpec` object.

Or are you thinking of other potential properties that might be defined later on?

bam_masterdata/datamodel/vocabulary_types.py
```diff
@@ -1,6 +1,223 @@
 from bam_masterdata.metadata.definitions import VocabularyTerm, VocabularyTypeDef
 from bam_masterdata.metadata.entities import VocabularyType
 
+# changes
```
JosePizarro3: you can delete these

bam_masterdata/datamodel/object_types.py
```diff
@@ -9281,6 +9339,93 @@ class MassSpec(Instrument):
...
+        code="IONIZATION_TYPE",
+        data_type="CONTROLLEDVOCABULARY",
+        vocabulary_code="IONIZATION_TYPE",
+        property_label="Ion source",
+        description="""Ionization Type//Ionenquelle""",
+        mandatory=False,
+        section="Technical Details",
+    )
+
+    chromatography = PropertyTypeAssignment(
```
JosePizarro3: if you defined LcSystem and GcSystem I wondered whether this is necessary or, better, whether what you need is a property of type OBJECT

bam_masterdata/datamodel/vocabulary_types.py
```diff
@@ -1,6 +1,223 @@
...
+        description="""Supercritical Fluid Chromatography//Überkritische Fluidchromatographie""",
+    )
+
+    ic = VocabularyTerm(
+        code="IC",
+        label="IC",
+        description="""Ion Chromatography//Ionenchromatographie""",
+    )
+
+    none = VocabularyTerm(
```
JosePizarro3: no need of NONE, this can be assigned when the property is not mandatory and nothing is stored

bam_masterdata/datamodel/vocabulary_types.py
```diff
@@ -1,6 +1,223 @@
...
+    )
+
+
+class MsIonPolarity(VocabularyType):
+    defs = VocabularyTypeDef(
+        code="MS_ION_POLARITY",
+        description="""Ion polarity//Ionenpolarität""",
+    )
+
+    pos = VocabularyTerm(
```
JosePizarro3: more verbose please, POSITIVE, NEGATIVE, POSITIVE_NEGATIVE

JosePizarro3 (commented):
Hi, thanks for the work, I have some minor comments + 2 larger ones:
1. Maybe it is a good idea to clarify the role of CHROMOGRAPHY_TYPE and the two objects (LC_SYSTEM, GC_SYSTEM) you want to define. It seems this could be better represented by a property of type OBJECT and a common object type called CHROMATOGRAPHY.
2. Would it be a good idea to move MS-specific definitions to its own subfolder, `.../datamodel/mass_spectrometry`? Similar to what we have in welding/.

bam_masterdata/datamodel/object_types.py
janlisec: Resolution {numeric, no unit} = m2 / (m2-m1) = value indicating if the peaks of two m/z measured at the same time can be clearly differentiated. We measure ion intensities (y) on a continuous scale for m/z (x). While mass precision is a parameter for the peak apex (is it at 123.4 or at 123.45), resolution is a parameter for peak width. Best example: consider compound CN. m would be 12.000 + 14.003 = 26.003. It has two isotopes, [13]CN = 13.007 + 14.003 = 27.010 and C[15]N = 12.000 + 15.000 = 27.000. We could measure both isotopic masses with sufficient precision independently but without sufficient resolution we would measure a single peak instead of two. 'approximate' is in there, because resolution is dependent on device type and mass, i.e. TOF-MS devices would have lower resolution for lower masses, Orbitrap devices would have higher resolution for lower masses.

bam_masterdata/datamodel/object_types.py
janlisec: I think we could avoid this formal strictness here, but I can make the changes if you like.

bam_masterdata/datamodel/object_types.py
janlisec: We can consider removing this property or (better) subsituting against the new property CHROMATOGRAPHY_TYPE.  MS_HYPHENATION_METHOD is basically redundant with CHROMATOGRAPHY_TYPE where the latter contains contr.vocab.

bam_masterdata/datamodel/object_types.py
janlisec: MS_SCAN_RATE is the time resolution of the device, either specified Hz [1/s] or sometimes in milli seconds per scan [ms].

bam_masterdata/datamodel/object_types.py
janlisec: I had long discussions with Caro on how to represent our intruments best. I would keep both systems apart because more system-type specific properties might be added in the future. Actually, we run these machines in combinations GC-MS or LC-MS, combining a separation device with one or several detector devices. I had originally specified INSTRUMENT.ANALYTICAL_SYSTEM to reflect this, but Caro suggested otherwise, stating that this was too broad.

bam_masterdata/datamodel/vocabulary_types.py
janlisec: ok

bam_masterdata/datamodel/object_types.py
janlisec: see above

bam_masterdata/datamodel/vocabulary_types.py
janlisec: Please keep it. It is good to be expicit here and allow to state that no separation method was used before injection to the MS.

bam_masterdata/datamodel/vocabulary_types.py
janlisec: ok :)

bam_masterdata/datamodel/object_types.py
janlisec: Thanks

janlisec:
> Hi, thanks for the work, I have some minor comments + 2 larger ones:
> 
> 1. Maybe it is a good idea to clarify the role of CHROMOGRAPHY_TYPE and the two objects (LC_SYSTEM, GC_SYSTEM) you want to define. It seems this could be better represented by a property of type OBJECT and a common object type called CHROMATOGRAPHY.
> 2. Would it be a good idea to move MS-specific definitions to its own subfolder, `.../datamodel/mass_spectrometry`? Similar to what we have in welding/.

[1] I would like to upload the device objects of our group so that users can link there measurements to these devices and other groups can see what we have got. We had originally a very broad INSTRUMENT.ANALYTICAL_SYSTEM, got more specific now and also could set up INSTURMENT.CHROMATOGRAPHY as an intermediate solution (but would potentially blow this up with type specific properties in the long run). All solutions have advantages and disadvantages... :/

[2] fine with me, especially when we can expect more specific property sets to appear in the future

--> once you made decisions on the above issues, I would modify the PR accordingly, correct?

JosePizarro3:
> > Hi, thanks for the work, I have some minor comments + 2 larger ones:
> > 
> > 1. Maybe it is a good idea to clarify the role of CHROMOGRAPHY_TYPE and the two objects (LC_SYSTEM, GC_SYSTEM) you want to define. It seems this could be better represented by a property of type OBJECT and a common object type called CHROMATOGRAPHY.
> > 2. Would it be a good idea to move MS-specific definitions to its own subfolder, `.../datamodel/mass_spectrometry`? Similar to what we have in welding/.
> 
> [1] I would like to upload the device objects of our group so that users can link there measurements to these devices and other groups can see what we have got. We had originally a very broad INSTRUMENT.ANALYTICAL_SYSTEM, got more specific now and also could set up INSTURMENT.CHROMATOGRAPHY as an intermediate solution (but would potentially blow this up with type specific properties in the long run). All solutions have advantages and disadvantages... :/
> 
> [2] fine with me, especially when we can expect more specific property sets to appear in the future
> 
> --> once you made decisions on the above issues, I would modify the PR accordingly, correct?

Hi @janlisec 

Sorry for the delay, I was out on vacation.

[1] What is your opinion? Indeed, controlled vocabularies vs objects is always the question 🙂  so whatever you feel more natural and for longer-term sustainability of the model

I will answer the comments now. Feel free to go resolving the comments when you tackle them. There are a few comments where you didn't comment, so please take a look into these so not to overlook them.

Thanks for the work, Jan, this is looking good, my comments are quite minor fussy details🙂

bam_masterdata/datamodel/vocabulary_types.py
JosePizarro3: I see, ok for me 👍🏻

bam_masterdata/datamodel/object_types.py
JosePizarro3: Ok, I understand. Then this design choice looks good for me (also answering your point [1] below).

bam_masterdata/datamodel/object_types.py
JosePizarro3: Ok, then you can remove it, but we need to check if there is data in the platform first. Let me check this (same as with MS_IONIZATION_MODE.

bam_masterdata/datamodel/object_types.py
JosePizarro3: I passed this explanation to chatgpt and:
```python
        ...
        description="""Approximate mass resolving power (m/Δm). Indicates the ability to separate two closely spaced m/z peaks; higher values mean better peak separation. Resolution depends on instrument type and measured mass.//Ungefähres Massenauflösungsvermögen (m/Δm). Beschreibt die Fähigkeit, zwei nahe beieinander liegende m/z-Peaks zu trennen; höhere Werte bedeuten eine bessere Peak-Trennung. Die Auflösung hängt vom Instrumententyp und der gemessenen Masse ab."""
        ...
```

bam_masterdata/datamodel/object_types.py
JosePizarro3: I understand you. The actual ideal solution would be having an unique property of type REAL and multivalued. Leave it as it is and do you mind adding an inline comment like:
```python
# TODO change to REAL + multivalued when implemented in openBIS 7
```

I can query and maintain a list of TODO from the Python files for later releases, so these comments will help me a lot in the future.

bam_masterdata/datamodel/object_types.py
JosePizarro3: Ok, I see. Then, I'd suggest clarifying/formalizing the units and adding an attribute `units` to this: https://bamresearch.github.io/bam-masterdata/howtos/object_types/#units-with-pint

This is a hacky solution I implemented (it only adds the units in the property label), but I hope it can be implemented in the openBIS side soon enough

bam_masterdata/datamodel/object_types.py
janlisec: done

bam_masterdata/datamodel/object_types.py
janlisec: done. description changed

bam_masterdata/datamodel/object_types.py
janlisec: As discussed on Teams: you can delete and reassign the 3 current occurences.

bam_masterdata/datamodel/object_types.py
janlisec: I thought keeping it on VARCHAR is wise to allow both commonly used definitions (Hz and ms) but switched to REAL now and defined the unit as Hz. Users/parser for formats using ms scan time need to convert to Hz or 1/s.

bam_masterdata/datamodel/object_types.py
janlisec: I added the comment to change to REAL later and made a formatting suggestion to the user in the description.

janlisec:
I believe I made all changes according to our discussions. Let me know if not. :)

JosePizarro3:
> I believe I made all changes according to our discussions. Let me know if not. :)

Thanks!

I will take a look on Monday or Tuesday, and rebase this branch (it got a bit messy because of other branches I merged). Everything looks good to me.

bam_masterdata/datamodel/object_types.py
JosePizarro3: Just for completeness and as explained in private: 

- This property was used only by 3 objects in the ELN
- I moved the information to EXPERIMENTAL_DESCRIPTION
- Then we can safely remove this property

bam_masterdata/datamodel/object_types.py
JosePizarro3: Same as with MS_IONIZATION_MODE comment:

- This property was used only by 3 objects in the ELN (all of them were GC)
- I moved the information to EXPERIMENTAL_DESCRIPTION
- Then we can safely remove this property

JosePizarro3:
Hi @janlisec @caro-demi , check the latest changes:
- I defined you as the CODEOWNERS of the MS model (I can also define only janlisec if you prefer so, as Caro here is giving support, but I'd let you decide)
- I moved the MS-specific definitions to their own subfolder

Once you reply to me, I will merge this and include the changes in openBIS. @caro-demi if you are curious, I can show you how that is done automatically in a quick chat.

janlisec:
Yes, Jose. Please merge and include so that we can proceed with uploading instruments using these object definitions. Thank you.

PR #306, opened 2026-06-01
287 include parent child relationships from objects already in the inventory

### LINKED ISSUE #287: Include parent-child relationships from objects already in the inventory
Important if we generate, e.g., experimental steps in the parser and we need some automated method to link with objects already in openBIS.

So far, the only option was to create objects and add_relationship between them at the same parsing level.

LucasZimm:
Entities:
- changed the add_relationship function to accept dictionarys
- added a function to get a unique code for a dictionary

Parser:
- added feature to create relationships with OpenBIS existing objects

Fixed:
- Last merge removed the @property from dataset attribute in object_type to decrease redundancy. This resulted in a small attribute change "dataset -> datasets"

### DIFF
--- a/bam_masterdata/cli/run_parser.py
+++ b/bam_masterdata/cli/run_parser.py
@@ -531,6 +531,7 @@ def run_parser_with_transactions(
                     logger.warning(f"Error occurred while fetching parent object: {e}")
             else:
                 parent_identifier = openbis_id_map[parent_id]
+                parent = openbis.get_object(parent_identifier)
             if isinstance(child_id, dict):
                 try:
                     child = openbis.get_object(**child_id)
--- a/bam_masterdata/cli/run_parser.py
+++ b/bam_masterdata/cli/run_parser.py
@@ -462,11 +462,11 @@ def run_parser_with_transactions(
     # TODO (May 2026) if later transactions support datasets change it to transaction.
     for object_instance in collection.attached_objects.values():
         try:
-            if object_instance.dataset != []:
+            if object_instance.datasets != []:
                 attached_datasets = openbis.new_dataset(
                     type="RAW_DATA",
                     sample=identifier,
-                    files=object_instance.dataset,
+                    files=object_instance.datasets,
                 )
 
                 attached_datasets.save()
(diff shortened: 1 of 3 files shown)


JosePizarro3:
~~@codex review~~ nevermind, this does not work in organization repos

bam_masterdata/metadata/entities.py
```diff
@@ -1155,13 +1168,13 @@ def remove(self, object_id: str = "") -> None:
             )
         del self.attached_objects[object_id]
 
-    def add_relationship(self, parent_id: str, child_id: str) -> str:
+    def add_relationship(self, parent_id: str | dict, child_id: str | dict) -> str:
```
JosePizarro3: Here is where adding a feature gets semantically tricky. This makes that using this function might not be trivial if someone sees "..._id" as a dictionary instead of a string.

Like now we are giving more flexibility to the inputs, I'd suggest to change `parent_id` / `child_id` to simply `parent` / `child`, and update the docstring of this function.

bam_masterdata/metadata/entities.py
```diff
@@ -1171,15 +1184,27 @@ def add_relationship(self, parent_id: str, child_id: str) -> str:
...
-        if (
-            parent_id not in self.attached_objects.keys()
-            or child_id not in self.attached_objects.keys()
-        ):
-            raise ValueError(
-                "Both `parent_id` and `child_id` must be assigned to objects attached to the collection."
+        for obj_id in (parent_id, child_id):
+            if not isinstance(obj_id, dict) and obj_id not in self.attached_objects:
+                raise ValueError(
+                    "Both `parent_id` and `child_id` must be assigned to objects attached to the collection or existing in the system."
```
JosePizarro3: I would change this a bit:
```python
                    "Both `parent` and `child` must be assigned to objects attached to the collection or being a dict readable by the system."
```

As up to here we do not know if the dict exist or not at all in openBIS, or what is its format.

bam_masterdata/cli/run_parser.py
```diff
@@ -504,23 +504,48 @@ def run_parser_with_transactions(
...
+            child.add_parents(parent)
+            rel_transaction.add(child)
 
-        child.add_parents(parent)
-        rel_transaction.add(child)
+            logger.info(
+                f"Prepared relationship: {child_identifier} -> parent {parent_identifier}"
+            )
+        else:
+            if isinstance(parent_id, dict):
```
JosePizarro3: maybe you can combine this logic in one inner, private function `_get_entity(entity, openbis, openbis_map.....)`

JosePizarro3 (commented):
Looks good, only a few comments for clarity.

Moreover, it would be nice if you could add testing of this implementation.

JosePizarro3 (approved):

PR #316, opened 2026-07-27
rework run parser test cases to work with the new transaction logic

### LINKED ISSUE #309: Rework Run_Parser Test Cases to work with the new transaction logic
### Test Cases
The current test cases are designed to work with the older run_parser function which is outdated and will be replaced by the run_parser_with_transactions function.

### Solution:
The run_parser's testfile needs to be updated for the newer function

LucasZimm:
Fixed:
- TypeError: cannot unpack non-iterable NoneType object for parser init by moving the error-handling for missing calls before the init

Changes:
- changed the tests to work with the new function
- added transaction mocks for testing

TODO:
- check if the logic for "test_run_parser_with_object_reference(...)" is outdated or something is missing in run_parser.py
(Line 119-120)

Closes #309

### DIFF
--- a/bam_masterdata/cli/run_parser.py
+++ b/bam_masterdata/cli/run_parser.py
@@ -31,27 +31,6 @@ def _parser_init(
     Returns:
         tuple: A tuple containing (collection, space, project, collection_openbis) if successful, None otherwise.
     """
-    # Ensure openbis is provided
-    if openbis is None:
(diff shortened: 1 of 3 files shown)


JosePizarro3 (approved):
Yeah, this makes sense. Did you realize because of the testing? 🙂

PR #320, opened 2026-07-30
SIMS new object types and vocabularies

### LINKED ISSUE #313: SIMS new object types and vocabularies
@elisabethjohn proposed [here](https://github.com/BAMresearch/masterdata-checker/issues/21) the following:

> I want to add:
> 
> class ExperimentalStep_SIMS(ObjectType)
> 
> class SIMSPrimaryIon(VocabularyType)
> 
> class SIMSLMIGSetting(VocabularyType)
> 
> class SIMSPolarity(VocabularyType)
> 
> class SIMSSputterSource(VocabularyType)
> 
> which are neccessary for ToF-SIMS experimental steps.
> 
> However, after adding the new data some tests failed.
> 
> [object_types.py](https://github.com/user-attachments/files/29733711/object_types.py)
> [vocabulary_types.py](https://github.com/user-attachments/files/29733712/vocabulary_types.py)

JosePizarro3:
Hi @elisabethjohn,

I've transfered the changes and before merging them into the datamodel. You can see the changes in the "Files changed" tab: https://github.com/BAMresearch/bam-masterdata/pull/320/changes

I have done some minor changes:

- I've fixed the names of `SIMSLMIGSetting` and `SIMSSputterSource` vocabulary terms to match their codes. Before, they were named `Bi1`, `Bi2`, `Bi3` (in the case of SIMSLMIGSetting) or `Positive`, `Negative` (for SIMSSputterSource), I guess from copy-pasting from other vocabulary.
- I have used `MS_ION_POLARITY instead of `SIMSPolarity`
- I've redefined `class ExperimentalStep_SIMS(ObjectType)` to `class SIMS(ExperimentalStep)`. Note that in Python the parenthesis points to inheritance, and as you are defining a subtype of ExperimentalStep, that's encoded by doing so. Furthermore, because you are inheriting, you do not need to copy-paste all the PropertyTypeAssignment that appear again in ExperimentalStep down in SIMS, e.g., name, show_in_project_overview, start_date,..., so you only need to define the properties specific for SIMS
- I've changed the `code` of the SIMS properties to match our current style and make it cleaner
- I've changed the label of `SIMS_CHARGE_COMPENSATION` to reflect better its boolean character

And can you address these small comments/questions:
1. Please, make sure that the order you used in SIMS is the one you want. This reflects how they will appear in the ELN at the end.
2. Right now, all other properties in `SIMS` that are not controlled vocabularies are marked as VARCHAR. They look to me like REAL properties. We can even define their units if so. Is that so?
3. Can you improve the `description` of SIMS a bit please? 🙂 
4. I like your addition of `customer`. I feel this is even more generic than SIMS. For now, I've added a `# TODO` comment so we do not forget to check it later so we can revisit if this property should be instead defined for `ExperimentalStep` and `SIMS`, and all other experimental steps, would inherit it.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,4 +9,4 @@
 # Owners of individual datamodel directories and its subdirectories
 /bam_masterdata/datamodel/welding/ @CagtayFabry
 /bam_masterdata/datamodel/creep_test/ @MarkusSchilling
-/bam_masterdata/datamodel/mass_spectrometry/ @janlisec @caro-demi
+/bam_masterdata/datamodel/mass_spectrometry/ @janlisec @caro-demi @elisabethjohn
--- a/bam_masterdata/datamodel/mass_spectrometry/object_types.py
+++ b/bam_masterdata/datamodel/mass_spectrometry/object_types.py
@@ -181,6 +181,7 @@ class SIMS(ExperimentalStep):
         generated_code_prefix="EXP.SIMS",
     )
 
+    # TODO check if this property should be moved to the ExperimentalStep type
     customer = PropertyTypeAssignment(
         code="CUSTOMER",
         data_type="VARCHAR",
--- a/bam_masterdata/datamodel/mass_spectrometry/object_types.py
+++ b/bam_masterdata/datamodel/mass_spectrometry/object_types.py
@@ -177,7 +177,7 @@ class GcSystem(Instrument):
 class SIMS(ExperimentalStep):
     defs = ObjectTypeDef(
         code="EXPERIMENTAL_STEP.SIMS",
-        description="""Experimental Step (ToF-SIMS)//Experimenteller Schritt (ToF-SIMS)""",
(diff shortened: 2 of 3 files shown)


JosePizarro3:
Ah, and forgot to say that I have defined your object type and vocabularies under the subfolder `...../datamodel/mass_spectrometry/`, next to Jan's definitions

elisabethjohn:
Thanks a lot, José!
Regarding your comments and questions:
1) sims_charge_compensation should be movend to the end of the section and it should be datatype REAL with unit V
2) Yes, you are right, they are REAL type properties, the units are in the description at the moment
3) more detailed description: Experimental step for a Time-of-Flight Secondary-Ion-Mass-Spectrometry (ToF-SIMS) analysis in spectrometry, depth profiling or imaging mode

JosePizarro3:
Ok great, thanks. I've made the changes. Please, check them out. If you are happy with them, we could include them in the `main` instance of openBIS. However, changing afterwards is a bit hard, although in your case it would be easier.

elisabethjohn:
I am happy with them :)

JosePizarro3:
Ok, perfect 🙂 

Thanks @elisabethjohn. The vocabularies and object type are now in the main instance. You should be able to activate them for your ELN/Inventory in the settings, so they can appear in the Dropdown menu.

As this is already in main, I am merging, hence closing.

PR #328, opened 2026-08-13
change to valid pint units or add ToDo

CagtayFabry:
I was trying a unit extraction script for pint (0.25.3) and noticed a few unit descriptions that are not convertiable to pint.Unit

I adjusted the unit where I could, but some are not defined or I am not sure what they mean (`rcf`)

### DIFF
--- a/bam_masterdata/datamodel/activities.py
+++ b/bam_masterdata/datamodel/activities.py
@@ -112,8 +112,8 @@ class Dcpd(ExperimentalStep):
     dcpd_temp_coeff = PropertyTypeAssignment(
         code="DCPD_TEMP_COEFF",
         data_type="REAL",
-        property_label="Temperature Coefficient of Resistivity [°C^-1]",
-        description="""Temperature Coefficient of Resistivity [°C^-1]//Temperaturkoeffizient der Resistivität [°C^-1]""",
+        property_label="Temperature Coefficient of Resistivity [K^-1]",
+        description="""Temperature Coefficient of Resistivity [K^-1]//Temperaturkoeffizient der Resistivität [K^-1]""",
         mandatory=False,
         section="Temperature Compensation",
     )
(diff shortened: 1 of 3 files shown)


bam_masterdata/datamodel/activities.py
```diff
@@ -112,8 +112,8 @@ class Dcpd(ExperimentalStep):
     dcpd_temp_coeff = PropertyTypeAssignment(
         code="DCPD_TEMP_COEFF",
         data_type="REAL",
-        property_label="Temperature Coefficient of Resistivity [°C^-1]",
```
CagtayFabry, line 115: `°C^-1` is a valid unit but it can cause problems when used for calculations 

I assume this coefficient is related to a change in temperature. In that case `K^-1` should be equivalent. Another pint-style alternative would be to use `delta_degC^-1` but that looks less clean to me.
docs: https://pint.readthedocs.io/en/stable/user/nonmult.html

JosePizarro3:
Thanks for spotting these. What about deleting the `[...]` ocurrences and just using the `units=` attribute in `PropertyTypeDef`?

CagtayFabry:
> Thanks for spotting these. What about deleting the `[...]` ocurrences and just using the `units=` attribute in `PropertyTypeDef`?

I was only parsing the entries to see if the units are valid for pint
do you have a script ready that can add the unit attributes? Maybe for a different PR, I could try to hack something together with AI if you want.. 🤖

JosePizarro3:
No, I don't have anything. There is a student who will handle this now when we start the transfer of the refactor from Miró.

Do you need this merged asap?

CagtayFabry:
> No, I don't have anything. There is a student who will handle this now when we start the transfer of the refactor from Miró.
> 
> Do you need this merged asap?

No it is not needed at all, I just found these and wanted to leave it here before I forget :)
I think in general it would be good practice to keep unit descriptions pint compatible when possible?

JosePizarro3:
> > No, I don't have anything. There is a student who will handle this now when we start the transfer of the refactor from Miró.
> > Do you need this merged asap?
> 
> No it is not needed at all, I just found these and wanted to leave it here before I forget :) I think in general it would be good practice to keep unit descriptions pint compatible when possible?

Ok, then, I'll keep an eye and implement these proper units when we refactor the model. I'll let Sagar know about this (he starts September 1st) and we will fix these inconsistencies.

I'll keep this open and tag it as "backlog".

Thanks again @CagtayFabry 🙂

PR #329, opened 2026-08-14
Objects and vocabularies for building materials

### LINKED ISSUE #324: Objects and vocabularies for building materials
Kevin shared some new definitions via email:

[object_type_SAMPLE.BUILDING_MATERIALS_SPECIMEN_FB7.1.xlsx](https://github.com/user-attachments/files/30734251/object_type_SAMPLE.BUILDING_MATERIALS_SPECIMEN_FB7.1.xlsx)

[object_type_SAMPLE.CONCRETE_MIXTURE_FB7.1.xlsx](https://github.com/user-attachments/files/30734253/object_type_SAMPLE.CONCRETE_MIXTURE_FB7.1.xlsx)

[vocabulary_BUILDING_MATERIALS_TEST_MACHINE_FB7.1.xlsx](https://github.com/user-attachments/files/30734255/vocabulary_BUILDING_MATERIALS_TEST_MACHINE_FB7.1.xlsx)

[vocabulary_SPECIMEN_GEOMETRY_TYPE_FB7.1.xlsx](https://github.com/user-attachments/files/30734256/vocabulary_SPECIMEN_GEOMETRY_TYPE_FB7.1.xlsx)

JosePizarro3:
Kevin, thanks for your submission.

If you want to see the definitions in Python (which are equivalent to Excel, just in a different format), you can check them out here: https://github.com/BAMresearch/bam-masterdata/pull/329/changes
For feedback back, it is enough to simply answer my points below.

I have some open points to clarify:

For SAMPLE.BUILDING_MATERIALS_SPECIMEN:

1. There is one missing controlled vocabulary definition in the Excels you sent me: SPECIMEN_ORIGIN
2. Can you clarify whether the SPECIMEN_GEOMETRY and SPECIMEN_ORIGIN are only for building materials? If so, we should rename them to something like BUILDING_MATERIALS_SPECIMEN_GEOMETRY/_ORIGIN.
3. There is a controlled vocabulary called BUILDING_MATERIAL_TYPE and you are defining the property being BUILDING_MATERIAL_GROUP. Do you mind clarifying this and deciding whether you want to call this "type" or "group"?
4. What is SPECIMEN_DIMENSIONS_SET_POINTS? Is it a real number?
5. I see "DENSITY" as a varchar. If this is instead a real number with units, can you clarify?
6. I will rename TEST_MACHINE to BUILDING_MATERIALS_TEST_MACHINE, in line with the associated controlled vocabulary code.
7. Also, why not calling "SAMPLE" instead of "SPECIMEN"? Or do you prefer to keep the term specimen all around?

For SAMPLE.CONCRETE_MIXTURE:

8. All properties you defined new are MULTILINE_VARCHAR. Do you mind clarifying what each of them mean? I think some of them are real numbers and some others might be simpler VARCHARs.
9. All of the properties are also with a Word Processor custom widget. Is this an intended behavior?

And in general:

10. Do you mind passing this to ChatGPT or any other AI tool and ask it to write better `description` for each of the objects/properties/vocabularies that you defined new? I can also do it, but won't be able to evaluate whether the machine is defining them properly or there might be some misconception.
11. I'd also like to give you ownership of these definitions, same as it happens with other applications like welding, creep testing, or mass spectrometry.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -10,3 +10,4 @@
 /bam_masterdata/datamodel/welding/ @CagtayFabry
 /bam_masterdata/datamodel/creep_test/ @MarkusSchilling
 /bam_masterdata/datamodel/mass_spectrometry/ @janlisec @caro-demi @elisabethjohn
+/bam_masterdata/datamodel/building_materials/ @DSSt-BM
--- a/bam_masterdata/datamodel/building_materials/__init__.py
+++ b/bam_masterdata/datamodel/building_materials/__init__.py
@@ -1,0 +1,1 @@
+"""Submodule containing type definitions related to building materials."""
--- a/bam_masterdata/datamodel/building_materials/object_types.py
+++ b/bam_masterdata/datamodel/building_materials/object_types.py
@@ -4,8 +4,8 @@ from bam_masterdata.metadata.definitions import ObjectTypeDef, PropertyTypeAssig
 
 class BuildingMaterialsSpecimen(Sample):
     defs = ObjectTypeDef(
-        code="SAMPLE.BUILDING_MATERIALS_SPECIMENT",
-        description="""A generic specimen used for building material tests, such as compressive strength tests.//Ein Standardprüfkörper, der für Baustoffprüfungen, wie beispielsweise Druckfestigkeitsprüfungen, verwendet wird.""",
+        code="SAMPLE.BUILDING_MATERIALS_SPECIMEN",
+        description="""A generic sample used for building material tests, such as compressive strength tests.//Ein Standardprüfkörper, der für Baustoffprüfungen, wie beispielsweise Druckfestigkeitsprüfungen, verwendet wird.""",
         generated_code_prefix="SAM.BMS",
     )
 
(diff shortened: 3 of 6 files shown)


DSSt-BM:
Hi José,

Thank you for the feedback. Regarding your comments...

* Definitions in Python:
    - (i) bam_masterdata/datamodel/building_materials/object_types.py has a wrong charater "T" at the end of the code string (line 7): class BuildingMaterialsSpecimen(Sample):
      defs = ObjectTypeDef(code="SAMPLE.BUILDING_MATERIALS_SPECIMENT", ...)
    - (ii) BuildingMaterialType already exists and should be updated:
            bam_masterdata/datamodel/building_materials/vocabularies.py
* (1) Within the attachment section of my e-mails from 10.07., 31.07., and 14.08.
* (2) I agree with your suggestions and have already updated the labels.
* (3) Since the BUILDING_MATERIAL_TYPE vocabulary already exists, I changed the name from "group" to "type".
* (4) This property is intended for describing all necessary dimension parameters. Due to the variaty of our geometries, this input should be flexible (e.g. "100 mm x 100 mm x 100   mm").
* (5) I have changed the data type to REAL.
* (6) I agree.
* (7) I have changed all names to "sample".
* (8) and (9) I was asked from OE 7.4 to submit this object. Since I have no additional information, I would like to keep these input fields as flexible as possible.
* (10) I have updated all the descriptions.

Please let me know if you have any further concerns.

JosePizarro3:
Perfect, thanks for the reply.

I will integrate the changes tomorrow (Wed 02.09), so you will see the new object types in the main instance. I will let you know once it is ready.

JosePizarro3:
Hi @DSSt-BM,

I have added your changes. 

Just two comments for some general organizational things:

1. Do you mind clarifying 7.4, or indicating someone from 7.4 to reach out to me to clarify the object CONCRETE_MIXTURE? All of them being MULTILINE_VARCHAR is not something I would typically approve, but given time constraints, it is fine for now. 
2. In case you want to submit new definitions, please, include your excel in the issue. For this, you do not need to print out all object types, but simply the ones you are defining new.

Thanks again and all the best :-)

PR #332, opened 2026-08-17
Refactor parsers to use class-based structure

### LINKED ISSUE #331: Rework run_parser.py with Class
Rework run_parser.py with a RunParsers Class to avoid duplicate code parts, visibility and better modifiability.

LucasZimm:
Refactored parsers to use class-based structure.

Closes #331.

### DIFF
--- a/bam_masterdata/cli/run_parser.py
+++ b/bam_masterdata/cli/run_parser.py
@@ -1,6 +1,6 @@
-from collections import Counter
-
-from pybis import Openbis
+import hashlib
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
 
 from bam_masterdata.logger import logger
 from bam_masterdata.metadata.entities import (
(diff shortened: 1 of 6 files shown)


bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-    if not collection_name:
-        logger.info(
-            "No Collection name specified. Attaching objects directly to Project."
-        )
-        collection_openbis = project
-    else:
-        if collection_name.upper() in [c.code for c in project.get_collections()]:
-            collection_openbis = space.get_collection(
-                f"/{space_name}/{project_name}/{collection_name}".upper()
+            raise ValueError(
```
JosePizarro3: When you use `logger.error` the raise error becomes inecessary. You just need to substitute this by `return None`.

I've seen it happens below, so I will mark all of them.

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-        else:
-            logger.info("Replacing collection code with uppercase and underscores.")
-            collection_openbis = openbis.new_collection(
-                code=collection_name.replace(" ", "_").upper(),
-                type=collection_type,
-                project=project,
+        # Ensure the space, project, and collection are set
+        if not project_name:
+            logger.error("The Project name must be specified for the parser to run.")
+            raise ValueError(
```
JosePizarro3: substitute raise for return None (see comment above)

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-                # Value is an ObjectType instance, construct the path
-                if not value.code:
-                    logger.warning(
-                        f"OBJECT reference for property '{key}' has no code, skipping"
+        # Ensure the files_parser is not empty
+        if not files_parser:
+            logger.error(
+                "No files or parsers to parse. Please provide valid file paths or contact an Admin to add missing parser."
+            )
+            raise ValueError(
```
JosePizarro3, line 51: substitute raise for return None (see comment above)

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+                "No files or parsers to parse. Please provide valid file paths or contact an Admin to add missing parser."
+            )
+        # Ensure collection_type is valid
+        if collection_type not in ["COLLECTION", "DEFAULT_EXPERIMENT"]:
+            logger.error(
+                f"Invalid collection_type '{collection_type}'. Must be either 'COLLECTION' or 'DEFAULT_EXPERIMENT'."
+            )
+            raise ValueError(
```
JosePizarro3, line 57: substitute raise for return None (see comment above)

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+        self._openbis_init()
+        self.parsing()
+
+    def _openbis_init(self):
+        self._openbis_space()
+        self._openbis_project()
+        self._openbis_collection()
+        self._bam_masterdata_collection()
+
+    def _openbis_space(self):
```
JosePizarro3: I would simplify this a bit and, instead of assigning `self.space_name`, I would pass it as an input here (and in `_openbis_init()`).

The reason: you are using the space_name only here, so no need to use memory space to store it as a class attribute. Just imagine what an user (in this case, us) would do: do we want to access the information like `RunParsers().space_name` or this is not so necessary?

JosePizarro3: More feedback: you can rename this for clarity `_get_space(self, space: str = "")`.

JosePizarro3: And final feedback: I would return `space` here, and instead store it in the self.attribute in `_openbis_init()`.

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+        try:
```
JosePizarro3, line 106: If you follow my advice of not assigning `self.space` here, you can do:
```python
space = None
try:
    space = self.openbis.get_space(space_name)
except Exception:
    pass

if space is None:
    username = self.openbis.username.upper()
    for openbis_space in self.openbis.get_spaces():
        if openbis_space.code.endswith(username):
             self.logger.warning(".......")
             return space

logger.error("No usable space for ....")
return space
```

See, I just used returns and loggers for clarifying the situation. Also, defined a tmp variable for clarity.

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-            object_id,
-            object_instance,
-            openbis,
-            collection,
-            openbis_id_map,
-            collection_name,
-            space_name,
-            project_name,
-        )
+    def _openbis_project(self):
```
JosePizarro3: I have the same feedback as with `_openbis_spaces()`:

1. Change name to `_get_project()`
2. Pass `project_name` as an input
3. Return `project`
4. Store it in a self attribute in the `_openbis_init()`: `self.project = self._get_project(project_name)`

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
 
-        # Check if object already exists in openBIS, and if so, notify and get for updating properties
-        if not object_instance.code:
-            if not collection_name:
-                object_openbis = openbis.new_object(
-                    type=object_instance.defs.code,
-                    space=space,
-                    project=project,
-                    props=obj_props,
+    def _openbis_collection(self):
```
JosePizarro3: This is where my feedback gets a bit uglier, hence the decision you have to make. If you follow my advices, in this function:

1. Rename it to `_get_collection()`
2. Pass `space_name`, `project_name`, and `collection_name` as inputs
3. Check `if self.project is None` before assignin `collection = self.project` in the first conditional
4. Pass also `collection_type` as input
5. Return `collection` and store it as a self attribute in `_openbis_init()`. In this case, you have to store it in a `self.collection_openbis` attribute, to not collide with our own self.collection.

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-                else f"/{space_name}/{project_name}/{collection_name}/{object_instance.code}"
-            )
+                self.collection_openbis = self.openbis.new_collection(
+                    code=self.collection_name.replace(" ", "_").upper(),
+                    type=self.collection_type,
+                    project=self.project,
+                )
+            self.collection_openbis.save()
+
+    def _bam_masterdata_collection(self):
```
JosePizarro3: No need for this function. Simply use `self.collection = CollectionType()` in the `_openbis_init()`

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+        # Create a bam_masterdata CollectionType instance for storing parsed results
+        self.collection = CollectionType()
+
+    def parsing(self):
+        # Iterate over each parser and its associated files and store them in `collection`
+        for parser, files in self.files_parser.items():
+            parser.parse(files, self.collection, logger=self.logger)
+
+        return self.collection, self.space, self.project, self.collection_openbis
```
JosePizarro3: mmm why do you return these? I guess you can return `self.collection`, but why the others?

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+            )
+        self.space_name = space_name
+        self.logger = logger
+        self.openbis = openbis
+        self.project_name = project_name
+        self.collection_name = collection_name
+        self.collection_type = collection_type
+        self.files_parser = files_parser
+        self._openbis_init()
+        self.parsing()
```
JosePizarro3: Mmm I think this should be moved to the first line in `def run(self)`, as it is a runtime executino, not an instantiation effect

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+
+        identifier = openbis_id_map[obj]
+        return openbis.get_object(identifier)
+
+    def _load_object_props(self, object_id, object_instance):
+        """
+        Load and map object properties, resolving OBJECT type references to openBIS identifiers.
+
+        Args:
+            object_id: The local identifier of the object.
```
JosePizarro3: Some of these docstrings are still old, as most of these are not used as input anymore (e.g., `openbis` now it is `self.openbis`, so no need to specify here)

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
+                    for obj_id, obj_inst in self.collection.attached_objects.items():
+                        if obj_inst is value and obj_id in self.openbis_id_map:
+                            referenced_identifier = self.openbis_id_map[obj_id]
+                            break
+                    if not referenced_identifier:
+                        # Construct identifier from the object's code
+                        # Assume it's in the same space/project as the current object
+                        if not self.collection.name:
+                            referenced_identifier = (
+                                f"/{self.space_name}/{self.project_name}/{value.code}"
```
JosePizarro3: please, note that if you avoid storing `space_name` and `project_name` as self attrs, these need to be inputs of the function (here and a couple of lines below)

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
                 f"Error saving dataset for files {attached_datasets.files}: {e}"
             )
 
-    rel_transaction = openbis.new_transaction()
+    def _identifier(self, object_instance):
+        original_code = object_instance.code
+        unique_code = self._make_unique_code(original_code, self.code_counter)
```
JosePizarro3: here is a bit of the complication of using self attrs.

For simplicity, I won't define `self.code_counter` and just use it as an input both here and in `def _identifier()`.

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
...
+        if unique_code != original_code:
+            self.logger.warning(
+                f"Duplicate local code {original_code} → renamed to {unique_code}"
+            )
+            object_instance.code = unique_code  # Update the code in the object instance
+        identifier = (
+            f"/{self.space_name}/{self.project_name}/{unique_code}"
```
JosePizarro3: same thing on passing these as inputs if you go with my design idea

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
...
-                )
+                try:
+                    object_openbis = self.openbis.get_object(identifier)
+                    object_openbis.set_props(obj_props)  # update properties
+                    logger.info(
+                        f"Object {identifier} already exists in openBIS, updating properties."
+                    )
+                except Exception:
+                    logger.info(
+                        f"Object with code {object_instance.code} does not exist in openBIS, creating new one."
```
JosePizarro3: Is this the correct message?

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
-            object_instance,
-            openbis,
-            collection,
-            openbis_id_map,
-            collection_name,
-            space_name,
-            project_name,
+            pass
+        parent_obj = self._get_entity(
+            self.openbis,
```
JosePizarro3: So these self attrs do not need to be passed as inputs (here and below)

bam_masterdata/cli/run_parser.py
```diff
@@ -9,515 +9,324 @@
...
                 )
-                if not collection_name:
-                    object_openbis = openbis.new_object(
-                        type=object_instance.defs.code,
-                        code=object_instance.code,
-                        space=space,
-                        project=project,
-                        props=obj_props,
-                    )
+                raise
```
JosePizarro3: return None

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
...
+            )
 
-        logger.info(
-            f"Prepared relationship: "
-            f"{child_obj.identifier} -> parent {parent_obj.identifier}"
-        )
+            identifier = self._identifier(object_instance)
+
+            try:
+                object = self.openbis.get_object(identifier)
```
JosePizarro3: watch out when using `object` here, as it is protected by python

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
...
-        parent_obj = _get_entity(
-            openbis,
-            openbis_id_map,
-            parent,
-            "parent",
-        )
-        child_obj = _get_entity(
+
+
+class RunParsersWithTransactions(RunParsers):
```
JosePizarro3, line 534: Looking very good.

My comments done for `RunParsers` stand here, so in case you change stuff in `RunParsers` you need to pass one last time through here

JosePizarro3 (commented):
Ok, cool. There are just some minor comments, but the design is more clear now, in my opinion.

I'd also like to, once we merge this, do some typing. I will create another issue and assign myself, and later assign you as reviewer.

bam_masterdata/cli/run_parser.py
```diff
@@ -527,68 +336,188 @@ def run_parser_with_transactions(
...
+            **kwargs,
         )
+        self.rel_transaction = self.openbis.new_transaction()
+        self.obj_transaction = self.openbis.new_transaction()
 
-        if parent_obj is None or child_obj is None:
-            continue
+    def run(self):
+        self.openbis_id_map = {}
+        self.code_counter = {}
```
CagtayFabry: I think using [`collections.Counter`](https://docs.python.org/3/library/collections.html#collections.Counter) instead of managing your own dict will also do what you implemented here

bam_masterdata/cli/run_parser.py
LucasZimm: I thought having it as self could be used in case we have multiple parsers so no overwriting happens.

JosePizarro3: mmm interesting, I haven't thought about that.

Ok, it sounds good to me.

LucasZimm:
The last push should fix the error from #350.

JosePizarro3:
@LucasZimm I've worked a bit more in the underlying logic and so. Please, have a detail check. 

I think I will merge this (after you reply with a thumbsup emoji), but we can talk in detail and I show you the changes next Monday.

This pull requests takes, indeed, care of #350 but also #333.

LucasZimm:
Looks good, you did change back the "raise ValueError" at the beginning, is it better?

JosePizarro3:
> Looks good, you did change back the "raise ValueError" at the beginning, is it better?

Yes. In some cases, we save some validations all the time. The situation with logger is that: an instance is still created and used later, while with raise, we want to stop the process.

If I told you otherwise before, it was my mistake.

PR #361, opened 2026-09-07
Add TOOL object type for FB 8.2 equipment workflow

### LINKED ISSUE #354: Add TOOL object type
We evaluated the existing BAM object types for the FB 8.2 equipment workflow and concluded that a dedicated `TOOL` object type is required.

The existing equipment types introduce mandatory properties that are not applicable to many tools. Using them would therefore require users to enter metadata that has no meaningful value for this use case.

We propose adding `TOOL` (`Tool//Werkzeug`) as a separate Object Type. The definition has already been tested on the BAM-provided development openBIS instance and can be provided as a sanitized Excel or Python contribution.

sieversjosua:
Existing equipment types require metadata that does not apply to many tools in the FB 8.2 workflow. This adds a dedicated `TOOL` object type based on the Excel definition prepared for #354.

The 13 properties cover identification, location, responsibility, calibration and tool purpose, grouped into General Information and Equipment Details. Name, serial number, equipment status and tool purpose are mandatory. The code prefix is `TOOL`; automatic code generation is disabled, as in the Excel. `TOOL_PURPOSE` defines `CUT`, `DRILL` and `SCAN`. Existing vocabulary references and the `PERSON.BAM` reference are reused where specified, with shared labels and descriptions aligned with the existing model. No existing type definitions are changed.

The Excel's `MODEL_NAME`, `BAM_INVENTORY_NUMBER` (text) and `EQUIPMENT_STATUS` codes are retained, alongside the calibration fields. The source Excel's `TOOL_WORK_OUTPUT` property and vocabulary are named `TOOL_PURPOSE` in this contribution. Existing development instances using the old code would need a separate migration; this PR only adds schema definitions. `BAM_LOCATION_COMPLETE` remains a reference to the instance-provided vocabulary; no location terms are included.

Validation on Python 3.12:
- Comparison with the source Excel, accounting for the purpose rename, prefix and section changes, and direct `MasterdataChecker` validation in `individual` mode passed.
- `ruff check` and `ruff format --check` passed for both changed files.
- `python -m pytest -q --ignore=tests/cli --ignore=tests/metadata/test_entities_dict.py tests`: 247 passed, 1 skipped.
- The full suite (`python -m pytest -sv tests`) could not collect the CLI tests because of the existing `"Openbis" | None` annotation error in `cli/run_parser.py`. The CLI entry point is affected by the same error, so the model checker was called directly.

Closes #354.

### DIFF
--- a/bam_masterdata/datamodel/object_types.py
+++ b/bam_masterdata/datamodel/object_types.py
@@ -1257,7 +1257,7 @@ class Tool(ObjectType):
     defs = ObjectTypeDef(
         code="TOOL",
         description="""Tool//Werkzeug""",
-        generated_code_prefix="S",
+        generated_code_prefix="TOOL",
         auto_generate_codes=False,
     )
 
@@ -1267,7 +1267,7 @@ class Tool(ObjectType):
         property_label="Name",
         description="""Name""",
         mandatory=True,
-        section="General",
+        section="General Information",
     )
 
     description = PropertyTypeAssignment(
@@ -1276,7 +1276,7 @@ class Tool(ObjectType):
         property_label="Description",
         description="""Short description and/or purpose//Kurzbeschreibung und/oder Zweck""",
         mandatory=False,
-        section="General",
+        section="General Information",
     )
 
     manufacturer = PropertyTypeAssignment(
@@ -1285,7 +1285,7 @@ class Tool(ObjectType):
         property_label="Manufacturer",
         description="""Manufacturer//Hersteller""",
         mandatory=False,
-        section="Equipment",
+        section="Equipment Details",
     )
 
     bam_location_complete = PropertyTypeAssignment(
@@ -1295,7 +1295,7 @@ class Tool(ObjectType):
         property_label="Complete BAM Location",
         description="""Complete BAM location (up to room level)//Komplette BAM-Ortsangabe (bis Raumlevel)""",
         mandatory=False,
-        section="Equipment",
+        section="Equipment Details",
     )
 
     bam_oe = PropertyTypeAssignment(
@@ -1305,7 +1305,7 @@ class Tool(ObjectType):
         property_label="BAM Organizational Entity",
         description="""BAM Organizational Entity//BAM Organisationseinheit (OE)""",
         mandatory=False,
-        section="Organisation",
+        section="Equipment Details",
     )
 
     model_name = PropertyTypeAssignment(
@@ -1314,7 +1314,7 @@ class Tool(ObjectType):
         property_label="Model Name",
         description="""Modellbezeichnung//Modellbezeichnung""",
         mandatory=False,
-        section="Equipment",
+        section="Equipment Details",
     )
 
     serial_number = PropertyTypeAssignment(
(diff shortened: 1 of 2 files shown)


bam_masterdata/datamodel/object_types.py
```diff
@@ -1253,6 +1253,137 @@ class AuxiliaryMaterial(ObjectType):
     )
 
 
+class Tool(ObjectType):
```
JosePizarro3, line 1256: For now, this is fine, but in a (very) near future, the inheritance of `Tool` wil go through something called `Equipment`. Just FYI.

With this, most of the properties you needed to define here for Tool will disappear, as they will be defined in other, upper classes. But it is good to keep this definition and keep `tool_work_out` as a controlled vocabulary of it, so we keep it as it is now.

bam_masterdata/datamodel/object_types.py
```diff
@@ -1253,6 +1253,137 @@ class AuxiliaryMaterial(ObjectType):
...
+    defs = ObjectTypeDef(
+        code="TOOL",
+        description="""Tool//Werkzeug""",
+        generated_code_prefix="S",
```
JosePizarro3:
```suggestion
        generated_code_prefix="TOOL",
```

bam_masterdata/datamodel/object_types.py
```diff
@@ -1253,6 +1253,137 @@ class AuxiliaryMaterial(ObjectType):
...
+    )
+
+    bam_oe = PropertyTypeAssignment(
+        code="BAM_OE",
+        data_type="CONTROLLEDVOCABULARY",
+        vocabulary_code="BAM_OE",
+        property_label="BAM Organizational Entity",
+        description="""BAM Organizational Entity//BAM Organisationseinheit (OE)""",
+        mandatory=False,
+        section="Organisation",
```
JosePizarro3: Please, note that if you have a section here in between `bam_location_complete` and the next property in Equipment, in the ELN these will appear separated.

JosePizarro3: If you want my opinion: I wouldn't over do the `section` of the properties. Just two or three max, and that's it. In your case, you can even put everything under the same section, or just define 2 different ones: "General Information" and "Equipment Details"

bam_masterdata/datamodel/object_types.py
```diff
@@ -1253,6 +1253,137 @@ class AuxiliaryMaterial(ObjectType):
...
+    last_calibration_date = PropertyTypeAssignment(
+        code="LAST_CALIBRATION_DATE",
+        data_type="DATE",
+        property_label="Last Calibration",
+        description="""Letzte Kalibrierung//Letzte Kalibrierung""",
+        mandatory=False,
+        section="Calibration",
+    )
+
+    tool_work_output = PropertyTypeAssignment(
```
JosePizarro3: is this the "purpose" of the tool? if so, simply call it `tool_purpose`, as well as the vocabulary_code and the defined vocabulary in vocabulary_types.py

JosePizarro3 (commented):
Very good, just some minor details.

Thanks for the submission @sieversjosua

sieversjosua:
Hi @JosePizarro3 thanks for the feedback, I just updated the changes accordingly! :)

JosePizarro3:
Many thanks, @sieversjosua !

I've added the definitions in the main instance, and I am merging this.

PR #365, opened 2026-09-09
Align TOOL properties with FB 8.2 requirements

### LINKED ISSUE #364: Align TOOL properties with the FB 8.2 requirements
Correct the TOOL definition introduced in #361: remove CALIBRATION_CERTIFICATION and LAST_CALIBRATION_DATE. Retain the generic optional DESCRIPTION and hide it in ELN edit views with show_in_edit_views=False.

The Python contribution preserves the four mandatory fields and TOOL_PURPOSE. Existing installations need the same assignment update; other types remain unchanged.

Follow-up to #354 and #361.

sieversjosua:
Remove the two calibration properties from TOOL. Keep the generic optional DESCRIPTION, hidden in ELN edit views with `show_in_edit_views=False`. The four mandatory fields and TOOL_PURPOSE remain unchanged.

Existing installations need the corresponding TOOL assignment update.

Validation: lint, format and model checks passed; targeted tests: 250 passed, 1 skipped. The full suite retains the same 4 failures and 51 setup errors as the baseline.

Closes #364. Follow-up to #361.

### DIFF
--- a/bam_masterdata/datamodel/object_types.py
+++ b/bam_masterdata/datamodel/object_types.py
@@ -1270,15 +1270,6 @@ class Tool(ObjectType):
         section="General Information",
     )
 
-    description = PropertyTypeAssignment(
-        code="DESCRIPTION",
-        data_type="MULTILINE_VARCHAR",
-        property_label="Description",
-        description="""Short description and/or purpose//Kurzbeschreibung und/oder Zweck""",
-        mandatory=False,
-        section="General Information",
-    )
-
     manufacturer = PropertyTypeAssignment(
         code="MANUFACTURER",
         data_type="VARCHAR",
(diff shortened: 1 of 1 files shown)


bam_masterdata/datamodel/object_types.py
```diff
@@ -1270,15 +1270,6 @@ class Tool(ObjectType):
         section="General Information",
     )
 
-    description = PropertyTypeAssignment(
```
JosePizarro3, line 1273: Here we have a challenge: `description` is a property which is generic enough to belong to all object types. My argument here, if it is not annoying for you is to keep it defined here, basically because it will eventually appear.

My recommendation is actually: 
1. Keep it defined
2. Add `show_in_edit_views=False` ([see this](https://github.com/BAMresearch/bam-masterdata/blob/main/bam_masterdata/metadata/definitions.py#L532)). This will toggle it out of the ELN and won't bother your users.

JosePizarro3 (commented):
Only one small thing 🙂 

Thanks!

JosePizarro3 (approved):

PR #368, opened 2026-09-14
Add project description and documentation object types

### LINKED ISSUE #367: Add project description and documentation object types
The FB 8.2 project workflow needs distinct object types for project descriptions and supporting documentation. These entries should be identifiable within existing collections, and documents should be attachable to the documentation entry.

Proposed Python contribution:

- `PROJECT_DESCRIPTION` with a required name and multiline description, using automatic code prefix `PRODES`.
- `PROJECT_DOCUMENTATION` with a required name, using automatic code prefix `PRODOC`.

The contribution adds the two object types using existing property definitions. Existing BAM types and assignments remain unchanged, so existing data and parsers do not require migration. The new types can be registered and used in the project workflow after review.

sieversjosua:
Add two object types for project entries:

- `PROJECT_DESCRIPTION`: required name and multiline description (`PRODES`).
- `PROJECT_DOCUMENTATION`: required name; project documents can be attached as datasets (`PRODOC`).

Existing types and collections remain unchanged.

Validation: lint, format and model checks passed; targeted tests: 250 passed, 1 skipped. The full suite retains the same 4 failures and 51 setup errors as the baseline.

Closes #367.


sieversjosua:
Hi @JosePizarro3 , one question: how will the process after merging work? When will it become available in Playground / Main? Or are there other steps necessary from my end? :)

JosePizarro3:
Hi @sieversjosua

So my message might have got lost, I repeat it here:

Do you mind clarifying a bit what is the intended use? I feel that the object type DOCUMENT and the own project description where objects are attached could be enough here.

Maybe, can you share a diagram where you plan to use this and argue why it is necessary.

sieversjosua:
The intention was to give each project a clear structure, separating its description from supporting documentation. I think you are right that this alone does not justify two new object types: the existing project description and DOCUMENT may be sufficient. We will check that the required document metadata is covered before revising this proposal.

JosePizarro3:
Thanks for the clarification. In that case, feel free to let me know.

Given that this might mean a stalled open pull request, I will cautionary give 1 month (until 23.10.2026) for the clarification, after what, if no response is received, I will close this with a final message.

Best regards 🙂

PR #369, opened 2026-09-14
Refactor code to move URLs into functions

### LINKED ISSUE #366: Move URL cli global variable inside functions
Right now, we define `URL` globally in cli.py. We need to move this into the specific functions to avoid collisions when using bam-masterdata in other apps.

@LucasZimm can you implement this, please?

LucasZimm:
Removed the URL global variable since it was only used in masterdata_sync.
Added an url variable to the masterdata_sync function.


LucasZimm:
Closes #366

bam_masterdata/cli/cli.py
JosePizarro3: Can you also add the same click.option in this CLI function and the if/else you added in the masterdata_sync but here?

JosePizarro3 (approved):
Thanks, just one minor thing! :)

JosePizarro3:
I need to release a version with a fix for this. I have pushed and will be merging now.

